PR #3, opened 2015-01-25
Use glib's GChecksum for md5/sha1

dequis:
This changes behavior slightly:

 - md5_init()/sha1_init() allocate a GChecksum
 - md5_finish()/sha1_finish() close and free() it
 - md5_digest_keep() was added (no sha1 equivalent needed)

And yes, glib has this concept of "closing" the GChecksum, which means
it can't be used anymore after g_checksum_get_digest().

jabber_cache_add() actually seems to need to do that to generate some
random-ish values, so i kept that working by adding a md5_digest_keep()
function that copies the GChecksum before it gets closed

----

Old PR for reference: https://github.com/dequis/bitlbee/pull/64

Copypasting relevant comments:

I haven't tested this other than 'it compiles'

Requires glib 2.16 (which we depend on already)

I'm trying to keep the code that uses these functions more or less working and minimize the changes outside, but dropping these interfaces and just using GChecksum directly in some parts probably won't hurt.

### DIFF
--- a/configure
+++ b/configure
@@ -52,7 +52,7 @@ pie=1
 arch=`uname -s`
 cpu=`uname -m`
 
-GLIB_MIN_VERSION=2.14
+GLIB_MIN_VERSION=2.16
 
 echo BitlBee configure
 
--- a/lib/md5.c
+++ b/lib/md5.c
@@ -1,273 +1,36 @@
-/*
- * MD5 hashing code copied from Lepton's crack <http://usuarios.lycos.es/reinob/>
- *
- * Adapted to be API-compatible with the previous (GPL-incompatible) code.
- */
-
-/*
- * This code implements the MD5 message-digest algorithm.
- * The algorithm is due to Ron Rivest.  This code was
- * written by Colin Plumb in 1993, no copyright is claimed.
- * This code is in the public domain; do with it what you wish.
- *
- * Equivalent code is available from RSA Data Security, Inc.
- * This code has been tested against that, and is equivalent,
- * except that you don't need to include two pages of legalese
- * with every copy.
- *
- * To compute the message digest of a chunk of bytes, declare an
- * MD5Context structure, pass it to MD5Init, call MD5Update as
- * needed on buffers full of bytes, and then call MD5Final, which
- * will fill a supplied 16-byte array with the digest.
- */
-
-#include <sys/types.h>
-#include <string.h>		/* for memcpy() */
-#include <stdio.h>
 #include "md5.h"
 
-static void md5_transform(uint32_t buf[4], uint32_t const in[16]);
-
-/*
- * Wrapper function for all-in-one MD5
- *
- * Bernardo Reino, aka Lepton.
- * 20021120
- */
-
-/* Turns out MD5 was designed for little-endian machines. If we're running
-   on a big-endian machines, we have to swap some bytes. Since detecting
-   endianness at compile time reliably seems pretty hard, let's do it at
-   run-time. It's not like we're going to checksum megabytes of data... */
-static uint32_t cvt32(uint32_t val)
+/* Creates a new GChecksum in ctx */
+void md5_init(md5_state_t *ctx)
 {
-	static int little_endian = -1;
-	
-	if (little_endian == -1)
-	{
-		little_endian = 1;
-		little_endian = *((char*) &little_endian);
-	}
-	
-	if (little_endian)
-		return val;
-	else
-		return (val >> 24) |
-		       ((val >> 8) & 0xff00) |
-		       ((val << 8) & 0xff0000) |
-		       (val << 24);
+	*ctx = g_checksum_new(G_CHECKSUM_MD5);
 }
 
-void md5_init(struct MD5Context *ctx)
+/* Wrapper for g_checksum_update */
+void md5_append(md5_state_t *ctx, const guint8 *buf, unsigned int len)
 {
-	ctx->buf[0] = 0x67452301;
-	ctx->buf[1] = 0xefcdab89;
-	ctx->buf[2] = 0x98badcfe;
-	ctx->buf[3] = 0x10325476;
-
-	ctx->bits[0] = 0;
-	ctx->bits[1] = 0;
+	g_checksum_update(*ctx, buf, len);
 }
 
-/*
- * Update context to reflect the concatenation of another buffer full
- * of bytes.
- */
-void md5_append(struct MD5Context *ctx, const md5_byte_t *buf,
-		unsigned int len)
+/* Wrapper for g_checksum_get_digest
+ * Also takes care of g_checksum_free(), since it can't be reused anyway
+ * (the GChecksum is closed after get_digest) */
+void md5_finish(md5_state_t *ctx, guint8 digest[MD5_HASH_SIZE])
 {
-	uint32_t t;
-
-	/* Update bitcount */
-
-	t = ctx->bits[0];
-	if ((ctx->bits[0] = t + ((uint32_t) len << 3)) < t)
-		ctx->bits[1]++;	/* Carry from low to high */
-	ctx->bits[1] += len >> 29;
-
-	t = (t >> 3) & 0x3f;	/* Bytes already in shsInfo->data */
-
-	/* Handle any leading odd-sized chunks */
-
-	if (t) {
-		unsigned char *p = (unsigned char *) ctx->in + t;
-
-		t = 64 - t;
-		if (len < t) {
-			memcpy(p, buf, len);
-			return;
-		}
-		memcpy(p, buf, t);
-		md5_transform(ctx->buf, (uint32_t *) ctx->in);
-		buf += t;
-		len -= t;
-	}
-	/* Process data in 64-byte chunks */
-
-	while (len >= 64) {
-		memcpy(ctx->in, buf, 64);
-		md5_transform(ctx->buf, (uint32_t *) ctx->in);
-		buf += 64;
-		len -= 64;
-	}
-
-	/* Handle any remaining bytes of data. */
-
-	memcpy(ctx->in, buf, len);
+	gsize digest_len = MD5_HASH_SIZE;
+	g_checksum_get_digest(*ctx, digest, &digest_len);
+	g_checksum_free(*ctx);
 }
 
-/*
- * Final wrapup - pad to 64-byte boundary with the bit pattern 
- * 1 0* (64-bit count of bits processed, MSB-first)
- */
-void md5_finish(struct MD5Context *ctx, md5_byte_t digest[16])
+/* Variant of md5_finish that copies the GChecksum
+ * and finishes that one instead of the original */
+void md5_digest_keep(md5_state_t *ctx, guint8 digest[MD5_HASH_SIZE])
 {
-	unsigned count;
-	unsigned char *p;
-
-	/* Compute number of bytes mod 64 */
-	count = (ctx->bits[0] >> 3) & 0x3F;
-
-	/* Set the first char of padding to 0x80.  This is safe since there is
-	   always at least one byte free */
-	p = ctx->in + count;
-	*p++ = 0x80;
-
-	/* Bytes of padding needed to make 64 bytes */
-	count = 64 - 1 - count;
-
-	/* Pad out to 56 mod 64 */
-	if (count < 8) {
-		/* Two lots of padding:  Pad the first block to 64 bytes */
-		memset(p, 0, count);
-		md5_transform(ctx->buf, (uint32_t *) ctx->in);
-
-		/* Now fill the next block with 56 bytes */
-		memset(ctx->in, 0, 56);
-	} else {
-		/* Pad block to 56 bytes */
-		memset(p, 0, count - 8);
-	}
-
-	/* Append length in bits and transform */
-	((uint32_t *) ctx->in)[14] = cvt32(ctx->bits[0]);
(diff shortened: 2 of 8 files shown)


jelmer:
This also needs updates for the md5 testsuite; #6 should make the travis build run make check too.

dequis:
The md5 testsuite passes without requiring changes (any code that uses md5_init and md5_finish in a simple way will work as-is), but adding make check to travis is a good idea too

dequis:
Merged the other PR and rebased, tests passing: https://travis-ci.org/bitlbee/bitlbee/builds/48285238

jelmer:
Ah, I guess the behaviour didn't change enough to affect the testsuite. Never mind that comment :)

FWIW Changes look good to me.

jgeboski:
While attempting to trial the patch:
```
GLib:ERROR:/var/tmp/portage/dev-libs/glib-2.42.1/work/glib-2.42.1/glib/gchecksum.c:1575:g_checksum_update: code should not be reached

Program received signal SIGABRT, Aborted.
```

Stacktrace:
```
#0  0xb7fdb92c in __kernel_vsyscall ()
#1  0xb7a6059f in raise () from /lib/libc.so.6
#2  0xb7a61fdc in abort () from /lib/libc.so.6
#3  0xb7eb675d in g_assertion_message (domain=domain@entry=0xb7f029ce "GLib", file=file@entry=0xb7f04b24 "/var/tmp/portage/dev-libs/glib-2.42.1/work/glib-2.42.1/glib/gchecksum.c", line=line@entry=1575, 
    func=func@entry=0xb7f04cd3 <__FUNCTION__.4578> "g_checksum_update", message=<optimized out>, message@entry=0x800dd018 "code should not be reached") at /var/tmp/portage/dev-libs/glib-2.42.1/work/glib-2.42.1/glib/gtestutils.c:2291
#4  0xb7eb6874 in g_assertion_message_expr (domain=domain@entry=0xb7f029ce "GLib", file=file@entry=0xb7f04b24 "/var/tmp/portage/dev-libs/glib-2.42.1/work/glib-2.42.1/glib/gchecksum.c", line=line@entry=1575, 
    func=func@entry=0xb7f04cd3 <__FUNCTION__.4578> "g_checksum_update", expr=expr@entry=0x0) at /var/tmp/portage/dev-libs/glib-2.42.1/work/glib-2.42.1/glib/gtestutils.c:2306
#5  0xb7e6c2dd in g_checksum_update (checksum=0x800d4138, data=0x8009e704 <next_id> "\a", length=4) at /var/tmp/portage/dev-libs/glib-2.42.1/work/glib-2.42.1/glib/gchecksum.c:1575
#6  0x8002ccf6 in md5_append (ctx=0xbfffe338, buf=0x8009e704 <next_id> "\a", len=4) at md5.c:12
#7  0x8004866e in jabber_cache_add (ic=0x800bb148, node=0x800bafe8, func=0x80044e16 <jabber_parse_roster>) at jabber_util.c:149
#8  0x80044df8 in jabber_get_roster (ic=0x800bb148) at iq.c:411
#9  0x80044d1f in jabber_pkt_bind_sess (ic=0x800bb148, node=0x800c0528, orig=0x800dd088) at iq.c:391
#10 0x800488ab in jabber_cache_handle_packet (ic=0x800bb148, node=0x800c0528) at jabber_util.c:225
#11 0x80043977 in jabber_pkt_iq (node=0x800c0528, data=0x800bb148) at iq.c:50
#12 0x800333ce in xt_handle (xt=0x800d9d70, node=0x800c0528, depth=0) at xmltree.c:193
#13 0x80033280 in xt_handle (xt=0x800d9d70, node=0x800bad10, depth=1) at xmltree.c:173
#14 0x80033243 in xt_handle (xt=0x800d9d70, node=0x0, depth=1) at xmltree.c:169
#15 0x8004297a in jabber_read_callback (data=0x800bb148, fd=11, cond=B_EV_IO_READ) at io.c:184
#16 0x800281b9 in b_event_passthrough (fd=11, event=2, data=0x800d2ef8) at events_libevent.c:147
#17 0xb7e0bff3 in event_process_active_single_queue () from /usr/lib/libevent-2.1.so.5
#18 0xb7e0cc54 in event_base_loop () from /usr/lib/libevent-2.1.so.5
#19 0xb7e0cf61 in event_base_dispatch () from /usr/lib/libevent-2.1.so.5
#20 0x80028099 in b_main_run () at events_libevent.c:86
#21 0x80025f55 in main (argc=3, argv=0xbfffee74) at unix.c:176
```

jgeboski:
Also, since all of the `GChecksum` stuff was introduced in glib-2.16, I would bump the version in `configure`.

dequis:
(disclaimer, i already said "I haven't tested this other than 'it compiles'")

```
(gdb) p ctx->gchecksum
$80 = (GChecksum *) 0x62500000d040
(gdb) p copy
$81 = (GChecksum *) 0x62500000d040
```

Seems like `g_checksum_copy()` uses this https://developer.gnome.org/glib/stable/glib-Memory-Slices.html

So... it may or may not allocate new memory for it.

>"The slice allocator can allocate blocks as small as two pointers, and unlike malloc(), it does not reserve extra space per block. For large block sizes, g_slice_new() and g_slice_alloc() will automatically delegate to the system malloc() implementation"

After inspecting a bit the original commit that added this functionality to jabber, 89d736a169cbff4520dcbb475aa7269b2cf4b837, i realized it's basically doing `md5(known_prefix + next_id)`, with an increasing `next_id`. Then I got enlightened and realized I could just keep that `known_prefix` to be always the same GChecksum, and close the one that I had copied instead, so b327cd81b99fe787c1d00cecf99175c694f18f87 happened

configure
```diff
@@ -52,7 +52,7 @@ pie=1
 arch=`uname -s`
 cpu=`uname -m`
 
-GLIB_MIN_VERSION=2.14
+GLIB_MIN_VERSION=2.16
```
Wilm0r: I was just thinking about this.. How about having a policy or so, and always allow this to be bumped up up to the glib version currently in Debian's oldest LTS release? (2.24 ATM)

Not to bump it up now without using its features, but if there's a good reason to do so, then do it.

Wilm0r:
So hmm, pull requests of untested code?

It looks reasonable now. Have you tested it after updating md5_digest_keep()?

jgeboski:
The following leak should be addressed:
```
==3824== 284 bytes in 1 blocks are definitely lost in loss record 80 of 87
==3824==    at 0x482B55C: malloc (in /usr/lib/valgrind/vgpreload_memcheck-x86-linux.so)
==3824==    by 0x48DF5B2: g_malloc (in /usr/lib/libglib-2.0.so.0.4200.1)
==3824==    by 0x48F7AFD: g_slice_alloc (in /usr/lib/libglib-2.0.so.0.4200.1)
==3824==    by 0x48F8035: g_slice_alloc0 (in /usr/lib/libglib-2.0.so.0.4200.1)
==3824==    by 0x48B7D9E: g_checksum_new (in /usr/lib/libglib-2.0.so.0.4200.1)
==3824==    by 0x134CBD: md5_init (md5.c:6)
==3824==    by 0x14F30E: jabber_generate_id_hash (jabber.c:269)
==3824==    by 0x14F2D6: jabber_connect (jabber.c:259)
==3824==    by 0x14EF9E: jabber_login (jabber.c:184)
==3824==    by 0x13D07C: account_on (account.c:362)
==3824==    by 0x128F78: cmd_account (root_commands.c:518)
==3824==    by 0x1282AE: cmd_identify_finish (root_commands.c:208)
```

dequis:
>jabber_generate_id_hash

Derp. Yeah I guess I should get rid of it on logout.

>So hmm, pull requests of untested code

Yup! I posted it mostly to get feedback on the way I adapted the existing md5/sha1 functions, while I was going through the whole list of pending patches.

Also not sure if it was mentioned, but `g_compute_checksum_for_data()` looks promising at first but it's useless for what we use md5/sha1 since it returns a string of hex characters.

>Have you tested it after updating md5_digest_keep()?

Yeah, although evidently I missed a valgrind run.

dequis:
Also might be worth changing:

* md5_byte_t is just a typedef for guchar. sha1 uses guint8.
* sha1_state_t and md5_state_t are structs that just hold a pointer to a GChecksum. Could get rid of those, use GChecksum directly where it's required.

Second one might result in changing every usage of these functions which is why I was avoiding it. But it's not like it hurts that much to change it.

jgeboski:
> md5_byte_t is just a typedef for guchar. sha1 uses guint8.

Yes.  Not to metion, any GLib based byte operations, such as with `GByteArray` use `guint8`.  Might as well keep it consistent.  (I think I commented on this matter elsewhere)

> sha1_state_t and md5_state_t are structs that just hold a pointer to a GChecksum. Could get rid of those, use GChecksum directly where it's required.

Or just typedef `sha1_state_t` and `md5_state_t` to `GChecksum` for compatibility.

configure
dequis: Relevant thread in my personal issue tracker: https://github.com/dequis/bitlbee/issues/25 - first comment has a summary of the versions. Copypasting it:

2.12 - centos 5 *(currently works by disabling twitter module)*
**2.14 - current bitlbee requirement**
2.16 - debian 5.0 (lenny) / ubuntu 8.04 lts
**2.20 - current libpurple requirement**
2.24 - debian 6.0 (oldstable, squeeze) / ubuntu 10.04 lts
2.26 - centos 6
2.32 - debian 7.0 (stable, wheezy) / ubuntu 12.04 lts
2.36 - centos 7
2.40 - ubuntu 14.04 lts
2.42 - debian 8.0 (testing, jessie)

With this commit we're keeping support for "debian 5.0 (lenny) / ubuntu 8.04 lts" as some sort of coincidence, mostly because glib 2.16 is a very interesting release. But I'm totally in favor of bumping it up to 2.24. Centos 6 is also important, plenty of users with that thing.

dequis:
Wrote this a while ago and forgot to submit: Pushed a refactor of md5. So I realized I'd have to do the exact same work in sha1. I think I'm going to create `lib/checksum.c` for the implementation and keep `lib/sha1.h` and `lib/md5.h`

Current progress: i turned sha1.c and md5.c into a bunch of macros. I think... this miiiiiight not be the best way to do it.

```
#define md5_init(ctx) checksum_init(ctx, G_CHECKSUM_MD5)
#define sha1_init(ctx) checksum_init(ctx, G_CHECKSUM_SHA1)
#define md5_finish(ctx, digest) checksum_finish(ctx, digest, MD5_HASH_SIZE)
#define sha1_finish(ctx, digest) checksum_finish(ctx, digest, SHA1_HASH_SIZE)
#define checksum_digest_keep(ctx, digest, len) checksum_finish(g_checksum_copy(ctx), digest, len)
#define md5_digest_keep(ctx, digest) checksum_digest_keep(ctx, digest, MD5_HASH_SIZE)
#define sha1_digest_keep(ctx, digest) checksum_digest_keep(ctx, digest, SHA1_HASH_SIZE)
#define checksum_append(ctx, buf, len) g_checksum_update(*ctx, buf, len)
#define sha1_append checksum_append
#define md5_append checksum_append
#define md5_free(x) g_checksum_free(*x)
```

jgeboski:
I'd be content with the macros, actually.  It prevents from creating a larger stack for no reason.

As for the macros, some of them need expression safety parenthesis:
```
#define checksum_append(ctx, buf, len) g_checksum_update(*(ctx), buf, len)
#define md5_free(x) g_checksum_free(*(x))
```

dequis:
Pushed macross. I'm still not convinced this is the sanest way to go, but if you say it's okay...

Wilm0r:
As much as I like the idea of not adding to the stack, maybe this can be achieved by having these functions use return?

Using macros feels a little too hacky to me. This is what one would use inline functiona for maybe.

dequis:
Bunch of random thoughts:

* Caring about the stack here feels like premature optimization. I'd rather not do anything for that reason. I just don't want to have 15 shitty tiny functions that look the same, that take up a bunch of space and don't do a lot. But I don't know if macros are the solution.
* I think these macros create a feeling of "where's all the code?". GCC is rather decent at printing errors with them, though
* Some functions are the exact same thing (`(md5|sha1)_append`), some are the same except for a  checksum type specific parameter (`(md5|sha1)_finish` with `(MD5|SHA1)_HASH_SIZE`). Macros seem great for that.
* I think I never asked: C99 stuff is okay right? We don't care about MSVC (lol) and seem to be compiling with `-std=gnu89`, gcc's current default, which is c89 with extensions that make it look a lot like c99. Personally I only care as long as it compiles in gcc and clang.
* c99 and gnu89 have slightly different inline function rules: http://www.greenend.org.uk/rjk/tech/inline.html (note to self, read this page more closely later)
* These md5_ and sha1_ functions should be usable when protocols are compiled as plugins, so can't make them 100% static inline

Wilm0r:
I did mean to mention the term premature optimisation somewhere, because yeah, we're definitely down those woods now.

I'd go for the first version of your patch that passes A) the valgrind and B) the "doesn't crash" test. :>

Re C99: I've avoided it for a long time (mostly declaring vars halfway a scope) but 1999 sure is a long time ago by now, so yeah, go for it.

dequis:
Reverted the macro stuff to the latest sane version, finished a few details and squashed [this tree][1] into 5627cee622e69bec8e9fe2fc060ec43bca177d8f. I consider it done.

[1]: https://github.com/dequis/bitlbee-new/commits/274a3661c965b423a66b141cbfc14fb10ade1d4e

jgeboski:
I have been running the original patch set without issue.  Looks good.

PR #4, opened 2015-01-25
Reorganize include files to avoid conflicts with other libs

dequis:
- Change all header includes to be relative to the project root
- Remove -I${includedir} from bitlbee.pc Cflags
- Install lib and protocols headers to their own directories. So now it is:

```
    /usr/include/bitlbee/*.h
    /usr/include/bitlbee/lib/*.h
    /usr/include/bitlbee/protocols/*.h
```

This breaks backwards compatibility of third party plugins, but now
they don't have to do ambiguous includes like `#include <proxy.h>`

This also fixes trac ticket [#1170][] - conflicts when libproxy and liboauth
are installed at the same time bitlbee is built, which the macports
project ran into several times.

------

Old PR: https://github.com/dequis/bitlbee/pull/57

Trac ticket: [#1170][] - "error: variable has incomplete type 'const struct oauth_service'" - conflict with liboauth in macports (also mentions conflict with libproxy)

Ticket in my personal tracker with @jgeboski asking for this: https://github.com/dequis/bitlbee/issues/21

[#1170]: http://bugs.bitlbee.org/bitlbee/ticket/1170

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@
 
 # Program variables
 objects = bitlbee.o dcc.o help.o ipc.o irc.o irc_im.o irc_channel.o irc_commands.o irc_send.o irc_user.o irc_util.o nick.o $(OTR_BI) query.o root_commands.o set.o storage.o $(STORAGE_OBJS) unix.o conf.o log.o
-headers = $(wildcard $(_SRCDIR_)*.h $(_SRCDIR_)lib/*.h $(_SRCDIR_)protocols/*.h)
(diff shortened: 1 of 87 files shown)


dequis:
Added "on hold" - let's get this merged before/during/after some other breaking changes, such as the reindentation which might happen at some point soon, or near a release. No point in merging it right now.

dequis:
Closing this, no need to keep it around, i'll just apply it whenever it becomes relevant

PR #5, opened 2015-01-25
purple: add topic and name_hint to groupchats

seirl:
Use imcb_chat_name_hint and imcb_chat_topic for the purple groupchats by using the PurpleConversation 'title' attribute.

### DIFF
--- a/protocols/purple/purple.c
+++ b/protocols/purple/purple.c
@@ -831,8 +831,11 @@ void prplcb_conv_new( PurpleConversation *conv )
 		struct groupchat *gc;
 		
 		gc = imcb_chat_new( ic, conv->name );
-		imcb_chat_name_hint( gc, conv->title );
-		imcb_chat_topic( gc, NULL, conv->title, 0 );
+		if( conv->title != NULL )
+		{
+			imcb_chat_name_hint( gc, conv->title );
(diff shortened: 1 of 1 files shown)


dequis:
:+1:

jgeboski:
Is it ever possible that `conv->title` is going to be `NULL`?  If so, wrap `imcb_chat_name_hint()` and `imcb_chat_topic()` in a conditional block.  Those functions do not gracefully handle `NULL` strings.

dequis:
Okay the patch is at +6 lines now, this means our possibilities to nitpick are duplicated. This is a bit hard though, all I can think is that the `!= NULL` is a bit redundant but I don't want to use that card yet and run out of nitpicks. @jgeboski any ideas?

jgeboski:
I completely agree with comparing to `NULL`.  That said, I don't think most of bitlbee compares against `NULL`, so it should probably be kept consistent.

dequis:
Are you saying we're out of nitpicks? :(

Wilm0r:
YES WE ARE!

Wilm0r:
(Though topics can be long so the long yet truncated life stories as channel names are going to be hilarious. But it beats #chat_000.

dequis:
Nitpick: missing a `)`

PR #7, opened 2015-01-25
random_bytes: Use /dev/urandom only, don't bother trying /dev/random

dequis:
Also abort() if there's no /dev/urandom

See http://www.2uo.de/myths-about-urandom/ for details.

----

Continues this old PR: https://github.com/dequis/bitlbee/pull/26 (I just finished adapting it to fix the issues that were pointed out there)

### DIFF
--- a/lib/misc.c
+++ b/lib/misc.c
@@ -413,69 +413,20 @@ signed int do_iconv( char *from_cs, char *to_cs, char *src, char *dst, size_t si
 		return outbuf - dst;
 }
 
-/* A pretty reliable random number generator. Tries to use the /dev/random
-   devices first, and falls back to the random number generator from libc
-   when it fails. Opens randomizer devices with O_NONBLOCK to make sure a
-   lack of entropy won't halt BitlBee. */
+/* A wrapper for /dev/urandom.
+ * If /dev/urandom is not present or not usable, it calls abort()
+ * to prevent bitlbee from working without a decent entropy source */
 void random_bytes( unsigned char *buf, int count )
 {
-	static int use_dev = -1;
-	
-	/* Actually this probing code isn't really necessary, is it? */
-	if( use_dev == -1 )
+	int fd;
+	if( ( ( fd = open( "/dev/urandom", O_RDONLY ) ) == -1 ) ||
+	    ( read( fd, buf, count ) == -1 ) )
 	{
-		if( access( "/dev/random", R_OK ) == 0 || access( "/dev/urandom", R_OK ) == 0 )
-			use_dev = 1;
-		else
-		{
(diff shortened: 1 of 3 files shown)


lib/misc.c
```diff
@@ -413,69 +413,22 @@ signed int do_iconv( char *from_cs, char *to_cs, char *src, char *dst, size_t si
...
-	
-	/* Actually this probing code isn't really necessary, is it? */
-	if( use_dev == -1 )
-	{
-		if( access( "/dev/random", R_OK ) == 0 || access( "/dev/urandom", R_OK ) == 0 )
-			use_dev = 1;
-		else
+	int fd;
+	if( ( fd = open( "/dev/urandom", O_RDONLY ) ) >= 0 ) {
+		if( read( fd, buf, count ) != -1 )
```
jgeboski: Merge this and the leading conditional into one?

lib/misc.c
```diff
@@ -413,69 +413,22 @@ signed int do_iconv( char *from_cs, char *to_cs, char *src, char *dst, size_t si
...
-		int i;
-		
-		/* Possibly the LSB of rand() isn't very random on some
-		   platforms. Seems okay on at least Linux and OSX though. */
-		for( i = 0; i < count; i ++ )
-			buf[i] = rand() & 0xff;
-	}
+
+	log_message( LOGLVL_ERROR, "/dev/urandom not present - aborting" );
+	abort();
```
jgeboski: I might be worth inverting the conditional above, and putting these lines in the conditional block.  It is typically cleaner to handle errors within the conditional that fails, and then have successful execution progress.

dequis: Agreed, although the "successful execution progress" here isn't very interesting. Either way yeah, looking at the old function it looks like those close() are handling failures when they are actually success

lib/misc.c
```diff
@@ -413,69 +413,20 @@ signed int do_iconv( char *from_cs, char *to_cs, char *src, char *dst, size_t si
...
+ * to prevent bitlbee from working without a decent entropy source */
 void random_bytes( unsigned char *buf, int count )
 {
-	static int use_dev = -1;
-	
-	/* Actually this probing code isn't really necessary, is it? */
-	if( use_dev == -1 )
+	int fd;
+	if( ( fd = open( "/dev/urandom", O_RDONLY ) ) == -1 ||
+	    ( read( fd, buf, count ) == -1 ) )
```
jgeboski: The second comparison to `-1` is wrapped in parenthesis, but not the first?

dequis: lol

jgeboski:
Looks good.

PR #8, opened 2015-01-25
twitter: implemented filter based group chats

jgeboski:
Filter group chats allow for the ability to read the tweets of select
users without actually following the users, and/or track keywords or
hashtags. A filter group chat can have multiple users, keywords, or
hashtags. These users, keywords, or hashtags can span multiple group
chats. This allows for rather robust filter organization.

The underlying structure for the filters is based on linked list, as
using the glib hash tables requires >= glib-2.16 for sanity. Since the
glib requirement of bitlbee is only 2.14, linked list are used in order
to prevent an overly complex implementation.

The idea for this patch was inspired by Artem Savkov's "Twitter search
channels" patch.

In order to use the filter group chats, a group chat must be added to
the twitter account. The channel room name is either follow:username,
track:keyword, and/or track:#hashtag. Multiple elements can be used by
separating each element by a semicolon.

### DIFF
--- a/protocols/twitter/twitter.c
+++ b/protocols/twitter/twitter.c
@@ -30,6 +30,215 @@
 #include "url.h"
 
 GSList *twitter_connections = NULL;
+
+static int twitter_filter_cmp(struct twitter_filter *tf1,
+                              struct twitter_filter *tf2)
+{
+	int i1 = 0;
+	int i2 = 0;
+	int i;
+
+	static const twitter_filter_type_t types[] = {
+		/* Order of the types */
+		TWITTER_FILTER_TYPE_FOLLOW,
+		TWITTER_FILTER_TYPE_TRACK
+	};
+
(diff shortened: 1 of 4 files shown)


jgeboski:
I have personally been running this patch for over a month on a bitlbee-3.2.2 instance without any issue.

jgeboski:
Documentation should probably be updated to reflect the changes.

dequis:
Considering this one reviewed enough from https://github.com/dequis/bitlbee/issues/54

Wilm0r:
ZOMG this is huge! \o/

I like the idea of it and the code I did see looked reasonable. One suggestion: Maybe adding an extra mode (mode=filtered or so?) to suppress the main stream because I would expect that some people with bizarre use cases want a filtered stream only?

Wilm0r:
(Oh yes, and I realise that I am commenting on a merged PR.)

PR #9, opened 2015-01-25
irc-channel: implemented a special mode for show_users

jgeboski:
This allows for users to be declared as being special, which does not
have any specific meaning. The meaning of being special is different
from protocol-to-protocol, which many protocols do not even implement.
This functionality is mainly geared towards a special user state which
only some protocols may actually need to define. For example, with the
third-party Steam plugin, this can be used for denoting a user which is
actively playing a game.

By default, this mode will not actually be used by any plugin. However,
it does default to the half-operator user mode.

### DIFF
--- a/protocols/bee.h
+++ b/protocols/bee.h
@@ -62,6 +62,7 @@ typedef enum
 	BEE_USER_AWAY = 4,      /* Compatibility with old OPT_AWAY flag */
 	BEE_USER_MOBILE = 8,    /* Compatibility with old OPT_MOBILE flag */
 	BEE_USER_LOCAL = 256,   /* Locally-added contacts (not in real contact list) */
+	BEE_USER_SPECIAL = 512, /* Denotes a user as being special */
 } bee_user_flags_t;
 
 typedef struct bee_user
--- a/irc.c
+++ b/irc.c
@@ -952,11 +952,11 @@ static char *set_eval_bw_compat( set_t *set, char *value )
 	             "channel setting instead.", set->key );
 	
 	if( strcmp( set->key, "away_devoice" ) == 0 && !bool2int( value ) )
-		val = "online,away";
+		val = "online,special%,away";
 	else if( strcmp( set->key, "show_offline" ) == 0 && bool2int( value ) )
-		val = "online@,away+,offline";
+		val = "online@,special%,away+,offline";
 	else
-		val = "online+,away";
+		val = "online+,special%,away";
 	
 	for( l = irc->channels; l; l = l->next )
 	{
--- a/irc.h
+++ b/irc.h
@@ -223,7 +223,7 @@ struct irc_control_channel
 	struct bee_group *group;
 	struct account *account;
 	struct prpl *protocol;
-	char modes[4];
+	char modes[5];
 };
 
 extern const struct bee_ui_funcs irc_ui_funcs;
--- a/irc_channel.c
+++ b/irc_channel.c
@@ -641,13 +641,13 @@ static gboolean control_channel_init( irc_channel_t *ic )
 	set_add( &ic->set, "protocol", NULL, set_eval_by_protocol, ic );
 	
 	/* When changing the default, also change it below. */
-	set_add( &ic->set, "show_users", "online+,away", set_eval_show_users, ic );
+	set_add( &ic->set, "show_users", "online+,special%,away", set_eval_show_users, ic );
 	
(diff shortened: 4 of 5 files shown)


jgeboski:
Documentation should probably be updated to reflect the changes.

protocols/bee.h
```diff
@@ -58,10 +58,11 @@ void bee_free( bee_t *b );
...
 {
-	BEE_USER_ONLINE = 1,    /* Compatibility with old OPT_LOGGED_IN flag */
-	BEE_USER_AWAY = 4,      /* Compatibility with old OPT_AWAY flag */
-	BEE_USER_MOBILE = 8,    /* Compatibility with old OPT_MOBILE flag */
-	BEE_USER_LOCAL = 256,   /* Locally-added contacts (not in real contact list) */
+	BEE_USER_ONLINE = 1,       /* Compatibility with old OPT_LOGGED_IN flag */
+	BEE_USER_AWAY = 4,         /* Compatibility with old OPT_AWAY flag */
+	BEE_USER_MOBILE = 8,       /* Compatibility with old OPT_MOBILE flag */
+	BEE_USER_LOCAL = 256,      /* Locally-added contacts (not in real contact list) */
+	BEE_USER_SPECIAL = 262144, /* Compatibility with old OPT_SPECIAL flag */
```
dequis: Why add compatibility stuff for something that didn't exist before?

also

>/* TODO(wilmer): Kill at least the OPT_ flags that have an equivalent here. */

jgeboski: Yeah, I contemplated this.  I will yank the `OPT_SPECIAL`.

Wilm0r:
Cool. Yeah, The %/halfop state is there, might as well use it. I think plumbing is in place elsewhere already that makes root do mode +h for these changes, right?

I guess an interesting followup would be to allow the user to change this state with protocols that support it, where this might have "special" effects on that IM network?

jgeboski:
> Cool. Yeah, The %/halfop state is there, might as well use it. I think plumbing is in place elsewhere already that makes root do mode +h for these changes, right?

For the most part, yes.  I have hacks in place to achieve this in the Steam plugin, it just seems as if other protocols could make use of such a feature.

> I guess an interesting followup would be to allow the user to change this state with protocols that support it, where this might have "special" effects on that IM network?

Yup!

dequis:
>I guess an interesting followup would be to allow the user to change this state with protocols that support it, where this might have "special" effects on that IM network?

As in show_users, or as in `/mode +h user`?

Also, heh, doing /mode of a user results in a message saying `Don't touch their modes`

jgeboski:
>As in show_users, or as in /mode +h user?

`show_users`.

protocols/bee.h
jgeboski: @Wilm0r is there any meaning behind these values?  I understand the compatibility flags, however, `BEE_USER_LOCAL` jumps straight to 256 (`0x100`), which is `OPT_TYPING`.  So, it would appear they're not completely "compatible."  Are any unused values up for grabs?

Wilm0r: So I left some space there as padding, since the LOCAL flag is kind of a different piece of info than online/away/mobile. Otherwise, nothing special. I don't think any of these values is inherited from Gaim anymore, but I could be wrong.

The OPT_ ones, yeah, I sure hope (didn't grep for it) they're not needed anymore anyway, definitely don't define new ones.

Wilm0r:
Ah no, I was actually talking about /mode +h. In the current implementation, the flag is set by the IM network, but there could be cases where it makes sense for the user to set it maybe. (And/or both.)

But no need to over-engineer since right now I can't think of a good example anyway. For now the steam mapping you have is pretty neat already.

dequis:
Way out of the scope of this PR but since you mentioned it:

Setting chanmodes would be more useful for groupchat moderation, and I guess that'd mean adding entries for MODE (and while we're at it, KICK, [relevant][1]) to the `irc_channel_funcs` struct which is defined separately for control channels and IM groupchats. So the control channel side of the implementation of the MODE handler could pass it to the protocol, maybe. Also blocking users with /ban would be neat.

[1]: https://github.com/jgeboski/bitlbee-facebook/issues/5

jgeboski:
This has been tested and works as expected without issue.

dequis:
:+1:

Merging soonish unless someone complains.

Wilm0r:
Sure, go for it.

PR #10, opened 2015-01-26
Twitter format strings

WPettersson:
Allow users to specify how tweets should be displayed

3 new settings are available to set how tweets are displayed:
 - twitter_format_string for normal tweets
 - retweet_format_string for retweets
 - reply_format_string for replies

For full documentation see the help files

### DIFF
--- a/protocols/twitter/twitter_lib.c
+++ b/protocols/twitter/twitter_lib.c
@@ -60,6 +60,7 @@ struct twitter_xml_status {
 	struct twitter_xml_user *user;
 	guint64 id, rt_id; /* Usually equal, with RTs id == *original* id */
 	guint64 reply_to;
+	struct twitter_xml_status *rt;
 };
 
 /**
@@ -85,6 +86,7 @@ static void txs_free(struct twitter_xml_status *txs)
 
 	g_free(txs->text);
 	txu_free(txs->user);
+	g_free(txs->rt);
 	g_free(txs);
 }
 
@@ -482,9 +484,9 @@ static struct twitter_xml_status *twitter_xt_get_status(const json_value *node)
 		struct twitter_xml_status *rtxs = twitter_xt_get_status(rt);
 		if (rtxs) {
 			g_free(txs->text);
-			txs->text = g_strdup_printf("RT @%s: %s", rtxs->user->screen_name, rtxs->text);
+			txs->text = g_strdup(rtxs->text);
 			txs->id = rtxs->id;
-			txs_free(rtxs);
+			txs->rt = rtxs;
 		}
 	} else if (entities) {
 		txs->text = expand_entities(txs->text, entities);
@@ -602,6 +604,49 @@ static gboolean twitter_xt_get_status_list(struct im_connection *ic, const json_
 	return TRUE;
 }
 
+/**
+ * Function to properly format a tweet as per the users configuration.
+ */
+static char *twitter_msg_get_text(struct im_connection *ic, int log_id, int reply_to,
+				struct twitter_xml_status *txs, const char *prefix) {
+	gchar * format = set_getstr(&ic->acc->set, "format_string");
+	GString * text = g_string_new(NULL);
+
+	gchar *c;
+	if (reply_to != -1)
+		format = set_getstr(&ic->acc->set, "reply_format_string");
+	if (txs->rt)
+		format = set_getstr(&ic->acc->set, "retweet_format_string");
+
+	for (c = format; *c ; c++) {
+		if (!(*c == '%' && *(c+1))) {
+			text = g_string_append_c(text, *c);
+			continue;
+		}
+		c++; // Move past the %
+		switch (*c) {
+			case 'i':
+				g_string_append_printf(text, "%02x", log_id);
+				break;
+			case 'r':
+				if (reply_to != -1) // In case someone does put %r in the wrong format_string
+				g_string_append_printf(text, "%02x", reply_to);
+				break;
+			case 't':
+				if (txs->rt) // In case someone does put %t in the wrong format_string
+					text = g_string_append(text, txs->rt->user->screen_name);
+				break;
+			case 'c':
+				text = g_string_append(text, txs->text);
+				break;
+			default:
+				text = g_string_append_c(text, *c);
+		}
+	}
+	text = g_string_prepend(text, prefix);
+	return g_string_free(text, FALSE);
+}
+
 /* Will log messages either way. Need to keep track of IDs for stream deduping.
    Plus, show_ids is on by default and I don't see why anyone would disable it. */
 static char *twitter_msg_add_id(struct im_connection *ic,
@@ -640,19 +685,7 @@ static char *twitter_msg_add_id(struct im_connection *ic,
 	if (g_strcasecmp(txs->user->screen_name, td->user) == 0)
 		td->log[td->log_id].id = txs->rt_id;
 	
-	if (set_getbool(&ic->acc->set, "show_ids")) {
-		if (reply_to != -1)
-			return g_strdup_printf("\002[\002%02x->%02x\002]\002 %s%s",
-			                       td->log_id, reply_to, prefix, txs->text);
-		else
-			return g_strdup_printf("\002[\002%02x\002]\002 %s%s",
-			                       td->log_id, prefix, txs->text);
-	} else {
-		if (*prefix)
-			return g_strconcat(prefix, txs->text, NULL);
-		else
-			return NULL;
-	}
+  return twitter_msg_get_text(ic, td->log_id, reply_to, txs, prefix);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


dequis:
Old PR: https://github.com/dequis/bitlbee/pull/63 (was already reviewed there)

dequis:
Okay testing this thing now.

You submitted the PR against master instead of develop. Doesn't matter a lot (i'll merge to develop either way), but it's got a trivial conflict with the other twitter patch that I had to fix.

Not exactly an issue with the patch, but the bold in those settings is a bit too subtle:



And this:

```
01:46 <@dx|ebooks> help set format_string
01:46 <@root> Type: string
01:46 <@root> Scope: account
01:46 <@root> Default: \002[\002%i\002]\002 %c
```

It says \002 literally, which might not make a lot of sense for most users. Maybe change to ^B and explain below?

I had to stop testing this patch because it got really scary:

protocols/twitter/twitter_lib.c
```diff
@@ -640,19 +685,7 @@ static char *twitter_msg_add_id(struct im_connection *ic,
...
-		else
-			return g_strdup_printf("\002[\002%02x\002]\002 %s%s",
-			                       td->log_id, prefix, txs->text);
-	} else {
-		if (*prefix)
-			return g_strconcat(prefix, txs->text, NULL);
-		else
-			return NULL;
-	}
+  return twitter_msg_get_text(ic, td->log_id, reply_to, txs, prefix);
```
dequis: also, nit, whitespace here (two spaces instead of one tab)

WPettersson: My bad, will fix.

WPettersson:
I see your point re. \002 in the documentation, but unsure on how to best go ahead. One choice is to go with "^B[^B%i^B]^B %c" and say that the ^B represents a toggling of the bold property and to see the documentation for your IRC channel for appropriate command. Is that the best way forward? It seems a bit clunky, but at the same time I can't work out a better way without removing the bold tags completely. And removing the bold tags completely I don't want to do as they're a good way of demonstrating that any regular IRC codes (including color codes) should work.

WPettersson:
I see your point re. \002 in the documentation, but unsure on how to best go ahead. One choice is to go with "^B[^B%i^B]^B %c" and say that the ^B represents a toggling of the bold property and to see the documentation for your IRC channel for appropriate command. Is that the best way forward? It seems a bit clunky, but at the same time I can't work out a better way without removing the bold tags completely. And removing the bold tags completely I don't want to do as they're a good way of demonstrating that any regular IRC codes (including color codes) should work.

WPettersson:
I see your point re. \002 in the documentation, but unsure on how to best go ahead. One choice is to go with "^B[^B%i^B]^B %c" and say that the ^B represents a toggling of the bold property and to see the documentation for your IRC channel for appropriate command. Is that the best way forward? It seems a bit clunky, but at the same time I can't work out a better way without removing the bold tags completely. And removing the bold tags completely I don't want to do as they're a good way of demonstrating that any regular IRC codes (including color codes) should work.

WPettersson:
I see your point re. \002 in the documentation, but unsure on how to best go ahead. One choice is to go with "^B[^B%i^B]^B %c" and say that the ^B represents a toggling of the bold property and to see the documentation for your IRC channel for appropriate command. Is that the best way forward? It seems a bit clunky, but at the same time I can't work out a better way without removing the bold tags completely. And removing the bold tags completely I don't want to do as they're a good way of demonstrating that any regular IRC codes (including color codes) should work.

dequis:
Actual non-nitpicky bug: %a (from the retweet format string) isn't being handled 

> and say that the ^B represents a toggling of the bold property and to see the documentation for your IRC channel for appropriate command. Is that the best way forward?

Sure. Also, you posted that three times for some reason. That's almost as scary as my twitter screenshot with "%c %c %c %c ..."

WPettersson:
Yeah sorry github seems to be playing up, it keeps giving "An error has occurred" messages even when things go through. Deleted duplicates. And yup, there's a complete disconnect between documentation and the code which caused %a to not work. I've fixed that up (a 2 character patch).
However, I can't duplicate your "%c %c %c ..." issue. Did you change format_string to something unusual? You also mentioned another twitter patch, is it on develop? There might be some conflict there.

dequis:
>However, I can't duplicate your "%c %c %c ..." issue. Did you change format_string to something unusual?

Sorry, that's exactly what i did. It's not an issue, it was just doing weird stuff while trying to break it.

>You also mentioned another twitter patch, is it on develop? There might be some conflict there

Yes, 73ee390abec21c2b724175d4a5c9cfc27aa85bcf, last commit of the develop branch atm. The conflict is trivial (keep both sides)

dequis:
I got some memory leaks reported by valgrind while testing this: http://dump.dequis.org/HeXzh.txt

These didn't appear when testing the current develop branch, but I think it might not be a problem with this branch. Not sure. Might be related to tweet entities (images, links, etc), and the tweets I received during that testing were kinda shitty.

protocols/twitter/twitter_lib.c
```diff
@@ -602,6 +604,49 @@ static gboolean twitter_xt_get_status_list(struct im_connection *ic, const json_
...
+		switch (*c) {
+			case 'i':
+				g_string_append_printf(text, "%02x", log_id);
+				break;
+			case 'r':
+				if (reply_to != -1) // In case someone does put %r in the wrong format_string
+				g_string_append_printf(text, "%02x", reply_to);
+				break;
+			case 'a':
+				if (txs->rt) // In case someone does put %t in the wrong format_string
```
dequis: reaching critical levels of nitpicking here, but `%t`

WPettersson:
Found the memory leak. Was g_free'ing the retweet stored inside a txs, instead of doing a txs_free. Patch incoming.

dequis:
Whoa, dude, you're... too efficient.

WPettersson:
It's a public holiday, have to keep my brain engaged somehow.

dequis:
Mergesquashed as ce402b20d82ec323e6bd5e306de934773590742d in the develop branch. Thanks!

(Also thanks for dealing with all our nitpicks so quickly, 10/10 A++++ would review again)

Wilm0r:
Soo... I do apologise for being the party pooper, but I'm most likely going to roll this back. Please don't take this personally! Code-wise from what I can tell it's good, and it's great that you provided documentation right away.

But the reason: Just because something can be made a setting, does not mean it *should* be a setting. Over the years I've had many requests and submissions like this one and I didn't take them, because although to the submitter apparently they're a useful feature, I question whether they would be useful for many others, to whom these three settings will just be noise among the list of account settings.

All substantial changes to the code that serve a very small audience, only mean extra risk for bugs and extra burden (more code to maintain) for the maintainers. IOW, bloat. Also, this change would be simple enough to just maintain in your own branch, should you really want to have an alternative format for your Twitter channel. Yes, that is a bit of a burden, but in this way the burden is carried by you, the person who wants this feature. (Also, you can keep it simple without making it a setting and adding all the format string processing.)

Having said that, while discussing this change @dequis mentioned an alternate idea of using gettext with an obscure local to you locale where you could keep your alternative format string instead. That would likely be a much leaner way to get the same functionality, so that's nice.

Sorry about that, but I think with the gettext idea, we can both get out of this happily (though I don't think anyone is working on adding gettext yet, even though the idea has been around for years)..

WPettersson:
That's not a problem at all for me. To be honest, I run a bunch of extra patches already on my bitlbee install so I'll probably keep using this as well, but it's up to you whether you include it. There's also an issue where this code "may not" play nice with UTF encoded format strings which I have been thinking about so there are valid concerns.

Wilm0r:
Hrm, interesting point, why do you think that would be a problem? If a % would be somewhere in a Unicode sequence, you mean? I don't think that should happen, any character part of a UTF-8 sequence should not be 7-bit-clean, see http://en.wikipedia.org/wiki/UTF-8#Description .

Just out of curiosity, what kind of format string changed are you running this with?

WPettersson:
Didn't realise the 7-bit-clean thing, but that makes sense. Ok, so no expected problem. And I just want the "RT %a" bit inside the square brackets. So [%i RT: @%a] would be my retweet format string. It makes more sense to me, I think of the bit inside the square brackets as "meta", whereas the text goes outside. I'm also considering making the square brackets bit a different colour as well.

Wilm0r:
Hmm yeah, use of colour is a rather personal thing.

The alternative format for RT's is interesting. "RT @$user: $text" is the default format and also what a retweet still looks like in the often-truncated field. But they look quite different in every Twitter GUI so definitely a good argument for making them look different in BitlBee, if the different version is clearer.

Though when people do manual retweets, they'll still usually look like the format above (unless this s one of those weird "$text (via @$user)" clients). Meh.

WPettersson:
Yeah part of the reason I want alternative ones is because I want to see the difference between someone actually retweeting via the API, or tweeting something that literally says "RT: @Blahblahblah said this"

dequis:
Soooo what do we do now? Revert? Adapt into something more generic? Pretend the feature doesn't exist and use it as an example of features that shouldn't exist? I like this last option because it means not doing anything right now.

Wilm0r:
Heh. I think I'll use this one as an excuse to learn how to do rollbacks in Git. So you yourself still don't need to do anything! \o/

PR #11, opened 2015-01-26
Add raw reply support

WPettersson:
By default, "reply" prepends the username of the tweet being replied to. This
adds support for a "rawreply" command which does not prepend this username.
This is useful for "replying" to your own tweet to maintain a chain of tweets
on a singular topic.

### DIFF
--- a/protocols/twitter/twitter.c
+++ b/protocols/twitter/twitter.c
@@ -673,6 +673,15 @@ static void twitter_handle_command(struct im_connection *ic, char *message)
 		message = new = g_strdup_printf("@%s %s", bu->handle, cmd[2]);
 		in_reply_to = id;
 		allow_post = TRUE;
+	} else if (g_strcasecmp(cmd[0], "rawreply") == 0 && cmd[1] && cmd[2]) {
+		id = twitter_message_id_from_command_arg(ic, cmd[1], NULL);
+		if (!id) {
+			twitter_log(ic, "Tweet `%s' does not exist", cmd[1]);
+			goto eof;
+		}
+		message += (cmd[2] - cmd[0]);
+		in_reply_to = id;
+		allow_post = TRUE;
 	} else if (g_strcasecmp(cmd[0], "post") == 0) {
 		message += 5;
 		allow_post = TRUE;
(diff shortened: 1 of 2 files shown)


dequis:
I have to admit I was FULLY AWARE of the existence of this patch in trac, and I kinda ignored it because I thought I could come up with something better UI-wise.

I didn't. So this is good enough. You win. This time.

Code looks okay.

protocols/twitter/twitter.c
```diff
@@ -673,6 +673,15 @@ static void twitter_handle_command(struct im_connection *ic, char *message)
 		message = new = g_strdup_printf("@%s %s", bu->handle, cmd[2]);
 		in_reply_to = id;
 		allow_post = TRUE;
+	} else if (g_strcasecmp(cmd[0], "rawreply") == 0 && cmd[1] && cmd[2]) {
+		id = twitter_message_id_from_command_arg(ic, cmd[1], NULL);
+		if (!id) {
+			twitter_log(ic, "Tweet `%s' does not exist", cmd[1]);
+			goto eof;
+		}
+		message += (cmd[2] - cmd[0]);
```
jgeboski: Shouldn't this duplicate the [functionality](https://github.com/bitlbee/bitlbee/blob/73ee390abec21c2b724175d4a5c9cfc27aa85bcf/protocols/twitter/twitter.c#L931) of `"reply"` via `message = cmd[2]`?  While it is essentially the same exact thing, it should be kept consistent, and avoids additional operations.

WPettersson: I'm not even sure what I was thinking when I wrote that (this patch was written 6 months ago). To be honest, I'm pretty sure the current implementation relies on split_command_parts not being re-written, so I'll push through a patch to just use "message = cmd[2]".

jgeboski:
Looks good.  Pending successful testing, I give it the green light for merging.

dequis:
Also looks good here after testing it. Only thing it's missing is adding the command to "help set commands", that'd be `doc/user-guide/commands.xml` where it says `<bitlbee-setting name="commands"...`.

(I could add that myself but you seem to be around)

dequis:
Mergesquashed as dfaa46d37c5a71e7d8c413c7c9714db854b8b559 in the develop branch. Thanks!

Wilm0r:
This should really have a suppress-attention-whore feature where .@ replies still aren't possible. >-)

Wilm0r:
Oh yes, and just to clarify: The reason I implemented it this way back then is because at the time including @username at the start of the Tweet was mandatory. That requirement was dropped a while later.

PR #13, opened 2015-01-29
irc_commands: implemented KICK support

jgeboski:
With similar commands being supported, such as INVITE, the KICK command
should be supported as well. The key motivation behind supporting KICK
is having for having a way to remove users from group chats. As of now,
there is no way for a bitlbee user to remove a user from a group chat.
With no current KICK implementation, it made using this command a prime
candidate for the UI side of this implementation. In addition, the KICK
command has been supported in the control channel as well. This is to
keep the INVITE/KICK pair consistent.

### DIFF
--- a/irc_channel.c
+++ b/irc_channel.c
@@ -625,6 +625,24 @@ static gboolean control_channel_invite( irc_channel_t *ic, irc_user_t *iu )
 	return TRUE;
 }
 
+static void control_channel_kick( irc_channel_t *ic, irc_user_t *iu, const char *msg )
+{
+	struct irc_control_channel *icc = ic->data;
+	bee_user_t *bu = iu->bu;
+	
+	if( bu == NULL )
+		return;
+	
+	if( icc->type != IRC_CC_TYPE_GROUP )
+	{
+		irc_send_num( ic->irc, 482, "%s :Kicks are only possible to fill_by=group channels", ic->name );
+		return;
+	}
+	
+	bu->ic->acc->prpl->remove_buddy( bu->ic, bu->handle,
+	                                 icc->group ? icc->group->name : NULL );
+}
+
 static char *set_eval_by_account( set_t *set, char *value );
 static char *set_eval_fill_by( set_t *set, char *value );
 static char *set_eval_by_group( set_t *set, char *value );
@@ -843,6 +861,7 @@ static const struct irc_channel_funcs control_channel_funcs = {
 	NULL,
 	NULL,
 	control_channel_invite,
+	control_channel_kick,
 	
 	control_channel_init,
 	control_channel_free,
--- a/irc_commands.c
+++ b/irc_commands.c
@@ -499,6 +499,30 @@ static void irc_cmd_invite( irc_t *irc, char **cmd )
 		irc_send_num( irc, 341, "%s %s", iu->nick, ic->name );
 }
 
+static void irc_cmd_kick( irc_t *irc, char **cmd )
+{
+	irc_channel_t *ic;
+	irc_user_t *iu;
+	
+	if( ( iu = irc_user_by_name( irc, cmd[2] ) ) == NULL )
+	{
+		irc_send_num( irc, 401, "%s :No such nick", cmd[2] );
+		return;
+	}
+	else if( ( ic = irc_channel_by_name( irc, cmd[1] ) ) == NULL )
+	{
+		irc_send_num( irc, 403, "%s :No such channel", cmd[1] );
+		return;
+	}
+	else if( !ic->f->kick )
+	{
+		irc_send_num( irc, 482, "%s :Can't kick people here", cmd[1] );
+		return;
+	}
+	
+	ic->f->kick( ic, iu, cmd[3] ? cmd[3] : NULL );
+}
+
 static void irc_cmd_userhost( irc_t *irc, char **cmd )
 {
 	int i;
@@ -745,6 +769,7 @@ static const command_t irc_commands[] = {
 	{ "ison",        1, irc_cmd_ison,        IRC_CMD_LOGGED_IN },
 	{ "watch",       1, irc_cmd_watch,       IRC_CMD_LOGGED_IN },
 	{ "invite",      2, irc_cmd_invite,      IRC_CMD_LOGGED_IN },
+	{ "kick",        2, irc_cmd_kick,        IRC_CMD_LOGGED_IN },
 	{ "topic",       1, irc_cmd_topic,       IRC_CMD_LOGGED_IN },
 	{ "oper",        2, irc_cmd_oper,        IRC_CMD_LOGGED_IN },
 	{ "list",        0, irc_cmd_list,        IRC_CMD_LOGGED_IN },
--- a/irc_im.c
+++ b/irc_im.c
@@ -971,6 +971,23 @@ static gboolean bee_irc_channel_chat_invite( irc_channel_t *ic, irc_user_t *iu )
 	return TRUE;
 }
 
+static void bee_irc_channel_chat_kick( irc_channel_t *ic, irc_user_t *iu, const char *msg )
+{
+	struct groupchat *c = ic->data;
+	bee_user_t *bu = iu->bu;
+	
+	if( ( c == NULL ) || ( bu == NULL ) )
+		return;
+	
+	if( !c->ic->acc->prpl->chat_kick )
+	{
+		irc_send_num( ic->irc, 482, "%s :IM protocol does not support room kicking", ic->name );
+		return;
+	}
+	
+	c->ic->acc->prpl->chat_kick( c, iu->bu->handle, msg );
+}
+
 static char *set_eval_room_account( set_t *set, char *value );
 static char *set_eval_chat_type( set_t *set, char *value );
 
@@ -1055,6 +1072,7 @@ const struct irc_channel_funcs irc_channel_im_chat_funcs = {
 	bee_irc_channel_chat_part,
 	bee_irc_channel_chat_topic,
 	bee_irc_channel_chat_invite,
+	bee_irc_channel_chat_kick,
 
 	bee_irc_channel_init,
 	bee_irc_channel_free,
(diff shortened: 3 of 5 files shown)


irc_channel.c
```diff
@@ -625,6 +625,24 @@ static gboolean control_channel_invite( irc_channel_t *ic, irc_user_t *iu )
...
+{
+	struct irc_control_channel *icc = ic->data;
+	bee_user_t *bu = iu->bu;
+	
+	if( bu == NULL )
+		return;
+	
+	if( icc->type != IRC_CC_TYPE_GROUP )
+	{
+		irc_send_num( ic->irc, 482, "%s :Kicks are only possible to fill_by=group channels", ic->name );
```
dequis: uhhh what? why?

Few people use fill_by=group because most protocols have shitty group support and I don't see why it should do that since all it does is remove_buddy, which removes a contact from the roster entirely. (also no idea why that one takes a group parameter)

irc_commands.c
```diff
@@ -499,6 +499,30 @@ static void irc_cmd_invite( irc_t *irc, char **cmd )
...
 }
 
+static void irc_cmd_kick( irc_t *irc, char **cmd )
+{
+	irc_channel_t *ic;
+	irc_user_t *iu;
+	
+	if( ( iu = irc_user_by_name( irc, cmd[2] ) ) == NULL )
+	{
+		/* KICK does not support ERR_NOSUCHNICK */
```
dequis: Send it anyway!

```
22:30 < seirl> jgeboski: what do you mean by "/* KICK does not support ERR_NOSUCHNICK */"?
22:32 < dx> << KICK #channel asdasdasdasdasdasd :
22:32 < dx> >> :some.server.net 401 dx asdasdasdasdasdasd :No such nick/channel
22:32 < seirl> yeah.
22:33 < dx> if the rfc said it's not supported, ignore it
22:34 < seirl> it's indeed not in the RFC
22:34 < seirl> but...
22:34 < seirl> oh, it's ERR_USERNOTINCHANNEL
22:35 < seirl> 441
22:35 < dx> ERR_NOSUCHNICK is often more accurate though
22:35 < seirl> yeah
22:37 < jgeboski> I was going by RFC 1459
22:37 < jgeboski> so I didn't support commas either
```

irc_im.c
```diff
@@ -969,6 +969,31 @@ static gboolean bee_irc_channel_chat_invite( irc_channel_t *ic, irc_user_t *iu )
...
+		irc_send_num( ic->irc, 482, "%s :Can't mix different IM networks in one groupchat", ic->name );
+		return;
+	}
+	else if( !c->ic->acc->prpl->chat_kick )
+	{
+		irc_send_num( ic->irc, 482, "%s :IM protocol does not support room kicking", ic->name );
+		return;
+	}
+	
+	/* TODO: Need more const goodness here, sigh */
```
dequis: wot?

jgeboski: The `prpl->chat_kick()` sends a non-constant string in order to keep consistent with the rest of the functions.  This is what other calls to these functions do.  Should it just avoid consistency, and be constant?

irc_im.c
```diff
@@ -969,6 +969,31 @@ static gboolean bee_irc_channel_chat_invite( irc_channel_t *ic, irc_user_t *iu )
...
+{
+	struct groupchat *c = ic->data;
+	bee_user_t *bu = iu->bu;
+	
+	if( bu == NULL )
+		return;
+	
+	if( c && ( iu->bu->ic != c->ic ) )
+	{
+		irc_send_num( ic->irc, 482, "%s :Can't mix different IM networks in one groupchat", ic->name );
```
dequis: This error message doesn't make sense for kicks. Also the numeric ERR_USERNOTINCHANNEL might make more sense in this particular case.

Is there no easy/efficient way to check if a nick is in a channel here? Not sure if it's okay to leave that to the protocol

irc_commands.c
jgeboski: I do not know if that is wise?

irc_channel.c
jgeboski: Same could be said for `control_channel_invite()`?

irc_im.c
dequis: IMO, yes, be constant. Turning the others into constants wouldn't hurt either.

Is this the reason you did the `g_strdup()` below? Because in that case, constants all the way.

irc_im.c
jgeboski: Good catch.  Will be addressed in the next revision of the patch.

irc_commands.c
dequis: Tested with 6 ircds over 8 networks (even ircnet and efnet), all of them behave like that. PRETTY SURE IT'S SAFE

irc_im.c
jgeboski: I was only doing what was [already](https://github.com/bitlbee/bitlbee/blob/1fa510944f3f95f1907422dc8f7cfe9238b99cd0/irc_im.c#L930]) done.  I guess the other functions are non-constant for Gaim compatibility?  In that case, I guess it does not hurt to be "inconsistent" for the sake of sanity?

irc_channel.c
dequis: oh. derp. huh.

https://github.com/bitlbee/bitlbee/commit/46d215d562f8e1aba2b24e2d1feab27337956d50

@Wilm0r hi

irc_im.c
dequis: Dammit jgeboski you sneaky copypaster. You're good at this.

>I guess the other functions are non-constant for Gaim compatibility?

I think keeping gaim compatibility shouldn't be a concern 10+ years after the fork happened :D. But I dunno. I'm going for "because nobody changed it"

>In that case, I guess it does not hurt to be "inconsistent" for the sake of sanity?

Yep

irc_channel.c
Wilm0r: What would you expect /invite to do in other cases? For fill_by=group, it can add a contact to a group so that the change persists through sessions. For anything else, the /invite can't really be mapped to IM space (can't add a FB account to a GTalk contact list, etc.).

protocols/nogaim.h
```diff
@@ -265,6 +265,12 @@ struct prpl {
 	/* If null, equivalent to handle_cmp( ic->acc->user, who ) */
 	gboolean (* handle_is_self) (struct im_connection *, const char *who);
 
+	/* This is called when the user uses the /kick IRC command.
+	 * - 'who' is a handle to kick
+	 * - 'message' is a kick message or NULL
+	 */
+	void (* chat_kick)	(struct groupchat *, char *who, char *message);
```
Wilm0r: I guess you're putting it here for slightly better binary compatibility of plugins? Otherwise, it's better sorted around chat_invite().

irc_channel.c
Wilm0r: Hrm, anyway. remove_buddy() takes a group param yeah, but if NULL it should remove the buddy altogether. That arg is inherited from Gaim really, I think. And since BitlBee doesn't support multiple groups per contact (because indeed mostly who cares about groups), it's effectively all the same.

jgeboski:
Updated.

dequis:
I dunno about the `fill_by=group` stuff but it's not relevant to this PR so i'll ignore that.

Looks okay, didn't review too deeply, didn't test.

dequis:
I did a quick implementation of this prpl in libpurple: e41cc4034741547c17ec4d24b9fdc50a8d2a5ee5. I dunno if it's rude to just push to the same branch you used here, so i made patch/kick+purple. Consider this an ad-hoc pull request, the commit is tiny.

It does work.

[offtopic] What wasn't so nice was that I tested it with the irc protocol of libpurple and that's fucking awful.

```
/join #dxtest124
<empty window for #dxtest124 gets created>
<window for #chat_000 gets created>

01:05 -!- dx|irc [dx@localhost.localdomain] has joined #chat_000
01:05 -!- ServerMode/#chat_000 [+t] by localhost.localdomain
01:05 [Users #chat_000]
01:05 [@dx|irc] [@root]
01:05 -!- Irssi: #chat_000: Total of 2 nicks [2 ops, 0 halfops, 0 voices, 0 normal]
01:05 -!- Topic for #chat_000: #dxtest124
01:05 -!- Topic set by root [root@localhost.localdomain] [Fri Jan 30 01:05:50 2015]
01:05 -!- dx [dx@irc] has joined #chat_000
01:05 -!- dx1 [dx1@irc] has joined #chat_000
01:05 -!- Irssi: Join to #chat_000 was synced in 0 secs
01:05 <@dx|irc> asd
01:06 < dxtest124> mode (+o dx1) by dx
01:06 < dxtest124> You have been kicked by dx1: (dx1)

<doesn't leave channel>
```

(`dx1` is the libpurple irc, `dx` is my usual irssi, latter gave ops to former, former kicked itself to test this command. `dxtest124` is a fake contact, not part of the contact list)

[/offtopic]

Awfulness aside, the kick itself worked.

Only one concern, how should the implementation report back an error to kick the user?

I'm thinking about something along the lines of `irc_send_num( ic->irc, 482, "%s :...", ic->name);` which belongs in irc_im.c in a bee_irc_chat_SOMETHING function? Then a bee_chat function to report errors back? I don't think this is done like that elsewhere.

Actually I think we something like that for channel join errors too, which are doing imcb_log and don't send a numeric, resulting empty windows like the one i mentioned before (the irc client thinks you joined, but no join messages or errors are received).

Maybe proper error reporting with numerics is something out of the scope of this?

dequis:
Merged this and my trivial purple implementation of it.

Decided to leave the error reporting stuff out, after all it's not really important for kicks.

Summary of out of scope issues that were mentioned here:

* `/invite` in control channels is for group management. This feature itself isn't documented as far as I know. May need reconsidering.
* Reconsider the "one group per contact" and the fact that add_buddy and remove_buddy take a group parameter
* Using irc through purple is a terrible experience (I already knew this, but found a few new issues here)
* We need something like imcb_error() that throws numerics on the irc side, but abstracted enough to a few generic kinds of errors. Maybe. Mostly for channel joins. I think it's irrelevant for kicks
* prpl functions need more consts, [particularly this][1]

[1]: https://github.com/bitlbee/bitlbee/blob/1fa510944f3f95f1907422dc8f7cfe9238b99cd0/irc_im.c#L930%5D

PR #14, opened 2015-02-02
Reindent everything to K&R style with tabs

dequis:
Parameters used:

    indent -i8 -kr -ts8 -ut -l100 **/*.{c,h}

Commit author set to `Indent <please@skip.me>` so that it's easier to skip while doing git blame.

-----

It was wilmer's idea, don't look at me like that (jgeboski also approves)

This matches the style used in the twitter module (5983eca9ac1187eb1407eece7d7ad72ef5e091b0). Also removes whitespace from empty lines which makes git happier.

## Diff stats

`170 files changed, 22111 insertions(+), 24197 deletions(-).`

**Note that viewing the diff through github may hang your browser for a few seconds**. Too much html. Just use the git commandline tools instead.

The patch file (with 3 lines of context) is **72506** lines long, slightly less than all of the code.

`wc -l **/*.{c,h}` before / after: **89727** vs **87641** - that's **2086** lines less \o/

## Things to do

- [x] Check that I didn't miss any file (anything that doesn't have .c or .h extension?)
    - `git ls-tree -r --name-only HEAD | egrep -v '\.(c|h|mock|po|pl|py|xml|xsl)$'` looks good
- [ ] Ensure that this *really really really* is the style we want to deal with for the next 100 years.
- [x] Wait for any patches by @seirl (telegram stuff, purple request input, utf8 channel names, etc)
    - `22:22 < seirl> dx: don't wait for me, i haven't started working on anything`
- [ ] Check for patches in trac (i'm thinking of carbons) and update them
    - [Full ticket list](http://bugs.bitlbee.org/bitlbee/query?status=!closed&keywords=~patch)
- [x] Write instructions on how to adapt any other out of tree patches to the new indent style, to minimize complaints
- [ ] Do something about ce402b20d82ec323e6bd5e306de934773590742d (also poke @WPettersson)


WPettersson:
I thought the format strings were being pulled out/not being merged in? Did I miss something?

dequis:
Well, we're supposed to revert that. It still hasn't happened. That's why "do something". If it's going to be reverted, it should happen before this patch gets merged.

Also poking you regarding any other patches you might have - if there's something else you want to get merged, send a PR now.

WPettersson:
I have one other patch, the view history/conversation behind a tweet one. However I'd rather wait until this is merged in, and then do mine afterwards. Mainly because with patches this size, it's easier to just commit this patch first, and then I can get my patch to fit in with that. It also means I can ensure my indentation and style are all good before sending a PR.

dequis:
Regarding style of the reindent: Sometimes it gets the alignment of multiline ifs wrong. Also i'm not sure if i like the way it aligns long lines such as the jabber_bs_abort function call below

```
			if (!(socks5_connect.ver == 5) ||
			    !(socks5_connect.cmdrep.cmd == 1) ||
			    !(socks5_connect.atyp == 3) || !(socks5_connect.addrlen == 40))
				return jabber_bs_abort(bt,
						       "Invalid SOCKS5 Connect message (addrlen=%d, ver=%d, cmd=%d, atyp=%d)",
						       socks5_connect.addrlen, socks5_connect.ver,
						       socks5_connect.cmdrep.cmd,
						       socks5_connect.atyp);
```

WPettersson:
```
@@ -371,10 +367,9 @@ static void twitter_oauth_start(struct im_connection *ic)
 	const char *url = set_getstr(&ic->acc->set, "base_url");
 
 	imcb_log(ic, "Requesting OAuth request token");
-	
+
 	if (!strstr(url, "twitter.com") && !strstr(url, "identi.ca"))
-		imcb_log(ic, "Warning: OAuth only works with identi.ca and "
-		             "Twitter.");
+		imcb_log(ic, "Warning: OAuth only works with identi.ca and " "Twitter.");```
```

Multi-line strings are being combined without strings being joined, could do with fixing.

```
@@ -571,10 +567,9 @@ static void twitter_login(account_t * acc)
 		return;
 	}
 
-	if (!strstr(url.host, "twitter.com") &&
-	    set_getbool(&ic->acc->set, "stream")) {
+	if (!strstr(url.host, "twitter.com") && set_getbool(&ic->acc->set, "stream")) {
 		imcb_error(ic, "Warning: The streaming API is only supported by Twitter, "
-		               "and you seem to be connecting to a different service.");
+			   "and you seem to be connecting to a different service.");
 	}
```

Again, multi-line strings are being aligned oddly.  (The second line should line up with "ic", it seems even github isn't displaying this one correctly. Check with command line tools for a better view).

`@@ -594,9 +589,9 @@ static void twitter_login(account_t * acc)` has the same issue.

```
--- a/protocols/twitter/twitter.h
+++ b/protocols/twitter/twitter.h
@@ -33,28 +33,25 @@
 #define debug( text... )
 #endif
 
-typedef enum
-{
-	TWITTER_HAVE_FRIENDS   = 0x00001,
-	TWITTER_MODE_ONE       = 0x00002,
-	TWITTER_MODE_MANY      = 0x00004,
-	TWITTER_MODE_CHAT      = 0x00008,
+typedef enum {
+	TWITTER_HAVE_FRIENDS = 0x00001,
+	TWITTER_MODE_ONE = 0x00002,
+	TWITTER_MODE_MANY = 0x00004,
+	TWITTER_MODE_CHAT = 0x00008,
 	TWITTER_DOING_TIMELINE = 0x10000,
-	TWITTER_GOT_TIMELINE   = 0x20000,
-	TWITTER_GOT_MENTIONS   = 0x40000,
+	TWITTER_GOT_TIMELINE = 0x20000,
+	TWITTER_GOT_MENTIONS = 0x40000,
 } twitter_flags_t;
```
enums are being "un-aligned"

Those are the only things I noticed. However I only went through the twitter code as that's all I'm really familiar with.

Wilm0r:
Though actually since the Twitter module is already indented with those rules pretty much, it can be skipped.

Also, the libyahoo2 and OSCAR code (apart from their c-files that glue it all to BitlBee) are mostly K&R. To limit the diff you could try excluding them from this change.

WPettersson:
The Twitter module does still have issues which need fixing with regards to indentation. You could even do each module one at a time, might be easier to handle.

Wilm0r:
Hrm, what kind of issues? The goal here is to clean up the excentric coding style that I used to use long ago and replace it with something more common. I'm pretty sure the Twitter module is mostly K&R already since it went through a full indent like this a few years ago already.

The goal is not to make this some kind of global reformatting, so *some* poor consistency here and there is not an issue.

WPettersson:
Oh, I see. I thought this was an attempt to get all of the code to match the style perfectly. If you don't mind minor things (mainly tabs/spaces and long lines) the Twitter module is fairly good.

Wilm0r:
Yep, no need to get over-ambitious - I trust human judgement more than an automated tool in the long run.

dequis:
Something else I almost forgot: I'd like to make block brackets mandatory, so instead of:

```
if (condition)
    single_statement_here();
```

It would be always:

```
if (condition) {
    single_statement_here();
}
```

There's a few particularly bad cases that have, say, an if inside a for inside another if. Don't remember where, I might have killed it.

I think GNU indent can't do this, but uncrustify does (under the codemod options). I'll have to write an equivalent uncrustify config file (the provided linux style example is a good starting point) and try to get it close this one.

Another benefit of uncrustify is that it seems to do the combination of tabs + spaces for alignment the way wilmer likes it.

dequis:
Regarding adapting out of tree patches, i think the least painful way would be (given `indent` as the indent commit and `indent^` as the previous one)

1. Apply patches against `indent^`, fix merge conflicts normally as needed, if any.
2. Run the exact same indent command. At the moment, that is:

    ```
    indent -i8 -kr -ts8 -ut -l100 **/*.{c,h}
    # TODO: the `**` is zsh syntax, replace it with find
    ```

3. Stage all the changes: `git add -u`
4. Create a temporary commit: `git commit -m "reindent temp commit"`
5. Rebase with `-X ours` against the indent commit:  `git rebase indent -X ours`
6. Review resulting diff and squash or make a new patch with `git diff indent`, whatever.

dequis:
`*poke*`

>I'd like to make block brackets mandatory

Opinion?

WPettersson:
I'd prefer to have mandatory block brackets. As you pointed out, they can sometimes avoid bugs with nested single-line if statements.

Wilm0r:
Yup, I said the same thing on IRC this morning, go for it!

dequis:
Okay that took a while but there it is. Threw away gnu indent. Some things that were mentioned before in this thread look much better now, uncrustify is definitely smarter. It's doing alignment with spaces, stuff isn't getting "un-aligned", etc.

It's two commits, one to add the config to `doc/uncrustify.cfg`, another to do the reindent itself.

I moved the migration notes here: https://wiki.bitlbee.org/Reindent

I decided to not exclude any files to keep it at just one command + one config file. Less moving parts. Or something.

For the curious, the interdiff between d7bccf6eece48e27e5fc8411038c18f52449ab32 (old indent) 2b675fe7b9686d7da2bf03de8ccbbe54f82276fa (new indent) is [here](http://dump.dequis.org/8OKiA.txt) (1.7mb txt). Diffstat is now `169 files changed, 25921 insertions(+), 25143 deletions(-)`, patch file line count is 80118. Also I just realized the line counts of the PR description are way off because I had other crap in my source tree (mainly the coverity `cov-int` directory). FWIW, i'm getting 68422 before, 69200 after.

And now I'd really like to get rid of this PR so please give it a quick look and let's get this shit merged.

WPettersson:
I'm still seeing multiline strings line up as if each separate part were a distinct variable (for instance, see protocols/twitter/twitter.c at lines 392/393). That one in particular can be fixed as it can be turned into a single line string, but one lines 601/602 we see the same thing. The strings are aligning where variables normally would. It's up to you whether this is desired or not, but I wasn't sure so wanted to point it out.

Other than that though, the twitter code looks fine.

Wilm0r:
But why reindent Twitter if it's already gone through a complete reindent? I'd avoid machine-reindenting code any part of the tree that doesn't need it. So that's all of protocols/twitter. Also, all of protocols/oscar and protocols/yahoo except yahoo.c (which is BitlBee glue to libyahoo2). Even oscar.c is inherited from Gaim and already mostly in the desired style.

WPettersson:
Twitter is missing bits. There's lots of one-line if statements without enclosing braces, and a few spacing issues. I'm not sure why you'd skip it, personally.

Wilm0r:
It depends on what the intended goal is. I suggested this just to get rid of the hard-to-follow (even for me by now) coding style I started with, not to rigorously get everything following *EXACTLY* one set of rules.

In the end I really trust proper human judgement on coding style more than any computer program, only reason we're using an automatic tool here now is because of course nobody would want to do this one-shot operation by hand.

So by excluding files that are in a workable style already, the extremely large diff by this change can be minimised, which is a win.

dequis:
> But why reindent Twitter if it's already gone through a complete reindent?

Mostly because of the new codemod options that make the `{ ... }` mandatory.

Also, quoting myself:

>I decided to not exclude any files to keep it at just one command + one config file. Less moving parts. Or something.

It was like 3am when i wrote this so let me clarify: I think it's saner to just run a single command that applies to the whole codebase and do all the changes automatically and in a reproducible way, instead of selectively excluding files or modifying them to make the diff only slightly smaller, which given its current size, it seems like a futile goal.

And I'd rather have something is trivial to apply to other out of tree patches than a small diff that requires a bunch of manual changes.

Wilm0r:
Sounds like overdoing it to me, but I'll buy the "trivial to apply" argument and don't care much anyway.

dequis:
OKAY THERE. IT'S OVER. FINALLY.

Wilm0r:
So, this seems to have screwed up comments:

<pre>
/********************************************************************\
  * BitlBee -- An IRC to other IM-networks gateway                     *
  *                                                                    *
  * Copyright 2002-2012 Wilmer van der Gaast and others                *
  \********************************************************************/
</pre>

In every single file. Why the hell did a code indenter reindent comments...

dequis:
Probably the settings for alignment (two spaces before it are weird) + the settings that tell it to reflow comments at all.

Resulting in... comments getting aligned to indentation levels but the body of the comment not getting reflowed. Yep.

PR #15, opened 2015-02-22
Filetransfer segfault fix!

dequis:
At least one of them!

Trac ticket: http://bugs.bitlbee.org/bitlbee/ticket/1122

Two commits here, one mostly uninteresting refactor I did a few weeks ago that turns copypasted code into functions, and the actual fix which i'll copy the commit log below:

-----

s5bytestream: fix segfault (cleanup before trying next streamhost)

This segfault happened when none of the available streamhosts can be connected to - or if at least one of them fails to connect.

Before this commit, it can be reproduced reliably by setting the `proxy` setting of the account to nonsense, for example, this is what i used:

    proxy.example.org,1.2.3.4,7777;proxy.example.com,173.194.42.65,80

`jabber_bs_recv_handshake_abort()` calls `jabber_bs_recv_handshake()`, which is supposed to restart the handshake with the next streamhost. And it replaced `bt->tf->watch_out`, which held an event ID, with a newer event ID. So the replaced event ID doesn't get removed, and it gets called again when its socket is closed by the timeout - and by the time that happens, the memory is free()'d already. Boom.

The patch is simple - created `jabber_bs_remove_events()` to cleanup those events, and use it before any code that expects to restart the cycle.

So basically the same as doing `b_event_remove(bt->tf->watch_out)`.

I hope there aren't more bugs like this in this code.

### DIFF
--- a/protocols/jabber/s5bytestream.c
+++ b/protocols/jabber/s5bytestream.c
@@ -252,6 +252,31 @@ gboolean jabber_bs_abort(struct bs_transfer *bt, char *format, ...)
 	}
 }
 
+void jabber_bs_remove_events(struct bs_transfer *bt)
+{
+	struct jabber_transfer *tf = bt->tf;
+
+	if (tf->watch_out) {
+		b_event_remove(tf->watch_out);
+		tf->watch_out = 0;
+	}
+
+	if (tf->watch_in) {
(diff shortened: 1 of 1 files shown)


Wilm0r:
Looks fine to me.

PR #16, opened 2015-02-22
Misc small patches

dequis:
* purple: cleanup, remove one usage of static local_bee
* storage_xml: replace hackish code with a GHashTableIter
* Don't hard-fail when building docs and dependencies are missing
* twitter_parse_id function, with better error handling than sscanf()

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -5,13 +5,13 @@ endif
 
 all:
 	# Only build the docs if this is a git tree
-	test ! '(' -d ../.git -o -d ../.bzr ')' || $(MAKE) -C user-guide
+	-test ! '(' -d ../.git -o -d ../.bzr ')' || $(MAKE) -C user-guide
 
 install:
 	mkdir -p $(DESTDIR)$(MANDIR)/man8/ $(DESTDIR)$(MANDIR)/man5/
 	$(INSTALL) -m 0644 $(_SRCDIR_)bitlbee.8 $(DESTDIR)$(MANDIR)/man8/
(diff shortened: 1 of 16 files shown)


dequis:
Added a few extra commits, coverity scan integration in .travis.yml (push to the coverity_scan branch to trigger it), and two commits fixing the lowest-hanging-fruit bugs reported by coverity.

doc/Makefile
```diff
@@ -5,13 +5,13 @@ endif
 
 all:
 	# Only build the docs if this is a git tree
-	test ! '(' -d ../.git -o -d ../.bzr ')' || $(MAKE) -C user-guide
+	-test ! '(' -d ../.git -o -d ../.bzr ')' || $(MAKE) -C user-guide
```
jelmer: This change probably means that the travis build will no longer catch breakage in the documentation build. Even if it does, other package builds can now accidentally skip the docs if this is the case.

Wouldn't a configure option (--doc=0) be better for this?

dequis:
Added --doc= configure option as suggested by jelmer, also improved the configure time checks for xmlto/xsltproc/a2x, now those aren't hard dependencies and the user is warned if they are missing.

I'd like to replace those deps with something entirely different at some point, like a python script or something. And, you know, I could even go ahead and write it in perl, which we already depend on (it's used in `doc/user-guide/Makefile` as a sed-on-steroids), so we don't add python as a dep.

dequis:
\*beep*

jelmer:
Doesn't look unreasonable to me.

dequis:
ha ha merged it anyway despite your not 100% certain phrasing

PR #17, opened 2015-02-27
Avoid crash on connecting to Jabber server

mcepl:
See https://bugzilla.redhat.com/show_bug.cgi?id=668190

Caused by our malfunctioning implementation of srv_lookup.


mcepl:
Sorry, it was made against `master` and it should be `develop`.

PR #18, opened 2015-02-27
Avoid crash on connecting to Jabber server

mcepl:
See https://bugzilla.redhat.com/show_bug.cgi?id=668190

Caused by our malfunctioning implementation of srv_lookup.

### DIFF
--- a/lib/misc.c
+++ b/lib/misc.c
@@ -34,6 +34,8 @@
 #include "nogaim.h"
 #include "base64.h"
 #include "md5.h"
+#include "misc.h"
+#include "ssl_client.h"
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


dequis:
Uh, the linked bug has a patch from 2011 which i can't quite tell what it does because it changes the indentation of the file. Is that one related?

How do I reproduce this issue? Grab centos5 box, build bitlbee, install, connect to any jabber server?

lib/misc.c
```diff
@@ -755,3 +842,24 @@ int truncate_utf8(char *string, int maxlen)
 	*end = '\0';
 	return end - string;
 }
+
+#ifdef TEST
+int main()
```
dequis: Test code doesn't belong here, there's a ./tests/ directory for that.

dequis:
Also I'm not really sure how that bug survived after 632627e1ead19fc7dc07effe441b2d543a675229

Possibly related issue from my personal issue tracker: https://github.com/dequis/bitlbee/issues/58 (only affects debian lenny, which is almost as old as centos 5)

dequis:
Oh uh, almost forgot, we dropped support for centos5 because of the glib version bump. Is this even relevant anymore?

mcepl:
> How do I reproduce this issue? Grab centos5 box, build bitlbee, install, connect to any jabber server?

Yes, that should be more or less it.

dequis:
Ehh, time to admit this one isn't going anywhere.

PR #19, opened 2015-02-28
Refactor oauth_params_del to fix use-after-free that i introduced

dequis:
Yeah ok that was dumb.

This is essentially just using a `data` variable instead of `l->data`, but i went ahead and cleaned up the function.


jgeboski:
Looks good.

PR #21, opened 2015-03-01
Purple : handle request_input

seirl:
(This change is not trivial because it needs to add some data in proto_data, which consisted solely of PurpleAccount. Hence, all the changes every time proto_data is casted to a PurpleAccount.)

### DIFF
--- a/protocols/purple/bpurple.h
+++ b/protocols/purple/bpurple.h
@@ -4,9 +4,14 @@
 #include <purple.h>
 #include <glib.h>
 
+#define PURPLE_REQUEST_HANDLE "purple_request"
+
 struct purple_data
 {
     PurpleAccount *account;
+
+    GHashTable *input_requests;
+    unsigned int next_request_id;
 };
 
 #endif /* !BPURPLE_H */
--- a/protocols/purple/bpurple.h
+++ b/protocols/purple/bpurple.h
@@ -1,0 +1,12 @@
+#ifndef BPURPLE_H
+# define BPURPLE_H
+
+#include <purple.h>
+#include <glib.h>
+
+struct purple_data
+{
+    PurpleAccount *account;
+};
+
+#endif /* !BPURPLE_H */
--- a/protocols/purple/purple.c
+++ b/protocols/purple/purple.c
@@ -39,6 +39,9 @@ static bee_t *local_bee;
 
 static char *set_eval_display_name(set_t *set, char *value);
 
+void purple_request_input_callback(unsigned int id, struct im_connection *ic,
+                                   const char *message, const char *who);
+
 struct im_connection *purple_ic_by_pa(PurpleAccount *pa)
 {
 	GSList *i;
@@ -310,6 +313,9 @@ static void purple_login(account_t *acc)
 
 	ic->proto_data = pd = g_new0(struct purple_data, 1);
 	pd->account = purple_account_new(acc->user, (char *) acc->prpl->data);
+	pd->input_requests = g_hash_table_new_full(g_int_hash, g_int_equal,
+	                                           g_free, g_free);
+	pd->next_request_id = 0;
 	purple_account_set_password(pd->account, acc->pass);
 	purple_sync_settings(acc, pd->account);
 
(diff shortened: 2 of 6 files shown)


dequis:
Noice.

dequis:
Thanks for splitting it in two commits this way! First commit looks good, and thanks to it, the second commit is much easier to read \o/

protocols/purple/bpurple.h
```diff
@@ -0,0 +1,17 @@
...
+
+#include <purple.h>
+#include <glib.h>
+
+struct purple_data
+{
+    PurpleAccount *account;
+
+    GHashTable *input_requests;
+    unsigned next_request_id;
```
dequis: unsigned what? either use `unsigned int` or `guint`

protocols/purple/purple.c
```diff
@@ -1040,9 +1067,57 @@ static void prplcb_request_test()
...
+	pd->next_request_id = (id + 1) % 10000;
+	ri->data_callback = ok_cb;
+	ri->user_data = user_data;
+	g_hash_table_insert(pd->input_requests, &id, ri);
+
+	char buddy[20] = { 0 };
+	sprintf(buddy, "%s_%d", PURPLE_REQUEST_HANDLE, id);
+
+	imcb_add_buddy(ic, buddy, NULL);
+	imcb_buddy_msg(ic, buddy, g_strdup(secondary), 0, 0);
```
dequis: leak in `g_strdup(secondary)`

protocols/purple/purple.c
```diff
@@ -1040,9 +1067,57 @@ static void prplcb_request_test()
...
+void purple_request_input_callback(unsigned id, struct im_connection *ic,
+                                   const char *message, const char *who)
+{
+	struct purple_data *pd = ic->proto_data;
+	struct request_input_data *ri = g_hash_table_lookup(pd->input_requests,
+	                                                    &id);
+
+	if (ri) {
+		void (*callback)(gpointer, const gchar *) =
+			(void (*)(gpointer, const gchar*)) ri->data_callback;
```
dequis: Uh, do a typedef for this function type?

protocols/purple/purple.c
```diff
@@ -1040,9 +1067,57 @@ static void prplcb_request_test()
...
+
+void* prplcb_request_input(const char *title, const char *primary,
+        const char *secondary, const char *default_value, gboolean multiline,
+        gboolean masked, gchar *hint, const char *ok_text, GCallback ok_cb,
+        const char *cancel_text, GCallback cancel_cb, PurpleAccount *account,
+        const char *who, PurpleConversation *conv, void *user_data)
+{
+	struct im_connection *ic = purple_ic_by_pa(account);
+	struct purple_data *pd = ic->proto_data;
+	unsigned id = pd->next_request_id;
```
dequis: another unsigned here

protocols/purple/purple.c
```diff
@@ -314,21 +325,28 @@ static void purple_login(account_t *acc)
...
+	purple_accounts_remove(pd->account);
 }
 
 static int purple_buddy_msg(struct im_connection *ic, char *who, char *message, int flags)
 {
 	PurpleConversation *conv;
+	struct purple_data *pd = ic->proto_data;
+
+	if (!strncmp(who, PURPLE_REQUEST_HANDLE, strlen(PURPLE_REQUEST_HANDLE))) {
+		unsigned request_id = atoi(who + strlen(PURPLE_REQUEST_HANDLE) + 1);
```
dequis: another one here too

dequis:
Agh dammit you submitted this PR against master. We should probably make that clearer...

Wilm0r:
I think the real problem is that due to the reindent, head and release have diverged much more than is reasonable/useful. Now I understand why that should have been postponed... Without that change, the base for submitted changes shouldn't be that important right?

dequis:
Nah, there's no "release" branch, that's what tags are for. The choice of "master" and "develop" and the assigned "meaning" that "master is more stableish than develop" is just by convention, or more precisely, a completely arbitrary distinction i made up (and that some other projects also follow).

I've already synced master with develop (as in, fast-forwarded a few commits because i considered those stable enough, if that means anything) two days ago. Before that, submitting the PR to the wrong branch would result in the ugly list of commits that https://github.com/bitlbee/bitlbee/pull/17 had, but this one didn't have that because master and develop were pointing at the same commit.

And, well, in this particular case, the base isn't very important - it just means that when i pushed the first two commits to the develop branch, this page didn't update (which, BTW, should have shown a merge conflict, since i modified the second commit to add a missing `g_free()` which seirl didn't do because he went afk)

So yeah, things that could be done to make this more intuitive:

* Change the default branch from master to develop (which means that the develop branch is the first thing users see when they view the repo, what people get when they clone, and most importantly the first option that appears when submitting a PR)
* Just have the latest code in the master branch, create a different branch for "stableish" code (decent name pending)
* Add a CONTRIBUTING.md file to remind people before they submit a PR: https://github.com/blog/1184-contributing-guidelines
* Write a bot that automatically closes pull requests submitted against master. I hate this solution but i've seen it done by other projects...

Why am I writing an essay here.

Also this PR is pretty much dead now that it needs to be closed and reopened against develop, so it doesn't matter if it's offtopic.

seirl:
Closing because of what dx said.

PR #22, opened 2015-03-02
purple: handle purple_request_input

seirl:
On develop this time. (see https://github.com/bitlbee/bitlbee/pull/21 for details)

*dx edit*: trac ticket: http://bugs.bitlbee.org/bitlbee/ticket/1178

### DIFF
--- a/protocols/purple/bpurple.h
+++ b/protocols/purple/bpurple.h
@@ -4,9 +4,14 @@
 #include <purple.h>
 #include <glib.h>
 
+#define PURPLE_REQUEST_HANDLE "purple_request"
+
 struct purple_data
 {
     PurpleAccount *account;
+
+    GHashTable *input_requests;
+    guint next_request_id;
 };
 
 #endif /* !BPURPLE_H */
--- a/protocols/purple/purple.c
+++ b/protocols/purple/purple.c
@@ -39,6 +39,17 @@ static bee_t *local_bee;
 
 static char *set_eval_display_name(set_t *set, char *value);
 
+void purple_request_input_callback(guint id, struct im_connection *ic,
+                                   const char *message, const char *who);
+
+/* purple_request_input specific stuff */
+typedef void (*ri_callback_t)(gpointer, const gchar *);
+
+struct request_input_data {
+	ri_callback_t data_callback;
+	void *user_data;
+};
+
 struct im_connection *purple_ic_by_pa(PurpleAccount *pa)
 {
 	GSList *i;
@@ -310,6 +321,9 @@ static void purple_login(account_t *acc)
 
 	ic->proto_data = pd = g_new0(struct purple_data, 1);
 	pd->account = purple_account_new(acc->user, (char *) acc->prpl->data);
+	pd->input_requests = g_hash_table_new_full(g_direct_hash, g_direct_equal,
+	                                           NULL, g_free);
+	pd->next_request_id = 0;
 	purple_account_set_password(pd->account, acc->pass);
 	purple_sync_settings(acc, pd->account);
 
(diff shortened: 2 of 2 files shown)


protocols/purple/purple.c
```diff
@@ -1055,9 +1068,57 @@ static void prplcb_request_test()
 }
 */
 
+struct request_input_data {
```
dequis: move this to the top!
also move that callback_t typedef outside of purple_request_input_callback
and use callback_t for the data_callback type instead of GCallback, save you some casting when calling it.

seirl:
Tested with telegram-purple, works fine, no leaks.

protocols/purple/purple.c
```diff
@@ -54,6 +57,13 @@ struct im_connection *purple_ic_by_pa(PurpleAccount *pa)
 	return NULL;
 }
 
+typedef void (*callback_t)(gpointer, const gchar *);
```
dequis: uh, any reason this is below purple_ic_by_pa?

protocols/purple/bpurple.h
```diff
@@ -4,9 +4,14 @@
...
 #include <glib.h>
 
+#define PURPLE_REQUEST_HANDLE "purple_request"
+
 struct purple_data
 {
     PurpleAccount *account;
+
+    GHashTable *input_requests;
+    unsigned int next_request_id;
```
dequis: use guint here, like in the other places?

dequis:
Nitpicks aside, looks good to me, but this is still pending some review by jgeboski, which should happen in a few days once he has time.

protocols/purple/purple.c
```diff
@@ -1055,9 +1076,50 @@ static void prplcb_request_test()
...
+	guint id = pd->next_request_id;
+	struct request_input_data *ri = g_new0(struct request_input_data, 1);
+
+	pd->next_request_id = (id + 1) % 10000;
+	ri->data_callback = (ri_callback_t) ok_cb;
+	ri->user_data = user_data;
+	g_hash_table_insert(pd->input_requests, GUINT_TO_POINTER(id), ri);
+
+	char buddy[20] = { 0 };
+	sprintf(buddy, "%s_%d", PURPLE_REQUEST_HANDLE, id);
```
jgeboski: You should probably just `g_strdup_printf()`.  Yes, it is "inefficient" by not working right on the stack, but only by a slight amount.  This will also allow for the full range of `guint` rather than just `0 - 9999`, granted that limit would probably never be reached.  Also, you should be using `%u`, not `%d`.

protocols/purple/purple.c
```diff
@@ -331,6 +346,12 @@ static int purple_buddy_msg(struct im_connection *ic, char *who, char *message,
 	PurpleConversation *conv;
 	struct purple_data *pd = ic->proto_data;
 
+	if (!strncmp(who, PURPLE_REQUEST_HANDLE, strlen(PURPLE_REQUEST_HANDLE))) {
+		guint request_id = atoi(who + strlen(PURPLE_REQUEST_HANDLE) + 1);
```
jgeboski: Why calculate the length of a static string repeatedly?  Just create another macro or a local static variable.

jgeboski:
Aside from my comments, and pending successful testing, it looks good.

protocols/purple/purple.c
dequis: As long as it's not a macro like `#define PURPLE_REQUEST_HANDLE_LENGTH strlen(PURPLE_REQUEST_HANDLE)`...

dequis:
Merged this commit with two fixes and an implementation of close_request here:

https://github.com/bitlbee/bitlbee/tree/purple_request_input

Quoting from irc:

>03:58 < dx> btw, there was a bug introduced by the change from strlen() to sizeof
>03:58 < dx> sizeof counts the null byte, strlen doesn't, so the comparison that checks if the nick start with purple_request failed

Also another minor bug was that there was one `&id` that should have been `GUINT_TO_POINTER(id)`

dequis:
Merged as 6a489929c121c20f7fad55fb6ec15c52643ee5ae with the two tiny fixes mentioned above.

The-Compiler:
I didn't actually test this until now, almost a year later. Works brilliantly, thanks @seirl @jgeboski @dequis! :+1:

Time to get rid of `telegram-cli`... finally!

PR #24, opened 2015-03-04
help: Increase read buffer size from 1100 to 2048

dequis:
This fixes "Error opening helpfile." that was introduced by the rawreply commit, which made the help string for "help set commands" larger than 1100 bytes, making the strstr(s, "\n%\n") condition fail

I have absolutely no idea why it was 1100 to begin with. It's been like that since the first revision. 4096 is arbitrary too, but it's nicer.

Needless to say, the help.c code needs to be reviewed more deeply, at least. Could also use that opportunity to get rid of the xmlto/xslt crap. Not going to do that right now, i don't want to add more bugs.

-------

PRing mostly to see what wilmer says, but I think asking for the reason of a magic number that hasn't been touched in 10+ years would be too much.

### DIFF
--- a/help.c
+++ b/help.c
@@ -29,7 +29,7 @@
 #undef read
 #undef write
 
-#define BUFSIZE 1100
+#define BUFSIZE 2048
 
 help_t *help_init(help_t **help, const char *helpfile)
 {


Wilm0r:
1100 was indeed quite arbitrary, but I did keep it small intentionally to keep help articles at a reasonable size (say, have them fit on a screen, no need for scrolling back). Growing the buffer slightly could be reasonable, but 4K is likely too large.

Re the xslt stuff: Why would you get rid of it? What's wrong with being able to generate docs in various formats from a single source? I think the main issue is that we're not really using the HTML docs in any useful way. It would've been neater if we could do a wiki export.

dequis:
>1100 was indeed quite arbitrary, but I did keep it small intentionally to keep help articles at a reasonable size (say, have them fit on a screen, no need for scrolling back).

lolk

>Growing the buffer slightly could be reasonable, but 4K is likely too large.

Going for 2048 then.

*(After writing that I decided to play 2048. Here I am, 30 minutes later, having achieved nothing at all, not even reaching 2048. Would not recommend.)*

>Why would you get rid of it? What's wrong with being able to generate docs in various formats from a single source? I think the main issue is that we're not really using the HTML docs in any useful way. It would've been neater if we could do a wiki export

I mostly want to get rid of the xmlto/xsltproc build time dependencies. Something like turning that into a simple python script. They are like 500mb of dependencies in a debian system unless you use --no-install-recommends. And yeah, it's not being used for anything significantly useful.

The generation scripts for the html output and other weirder formats (pdf/ps) can be kept, i just don't want to make it a requirement for all users who compile it.

jelmer:
They're only 500mb of dependencies if you install the recommends, too. If you don't produce the pdf and ps versions, you can drop the xmlto dependency entirely.

dequis:
pdf and ps versions are disabled, actually:

```
all: user-guide.txt user-guide.html help.txt # user-guide.pdf user-guide.ps user-guide.rtf
```

Note the `#`

xmlto is used for `%.txt`

```
%.txt: %.db.xml
	xmlto --skip-validation txt $<
	mv $*.db.txt $@
```

Which, now that I re-read these rules, it only seems to be for `user-guide.txt`

dequis:
Uhhh i'm going to go ahead and just merge this to get it out of the way since the patch itself is trivial. 

We can continue the doc refactoring discussion elsewhere (or in this same thread, if you want)

PR #25, opened 2015-03-06
twitter: Fix twitter_parse_id to accept 00 as a valid tweet

dequis:
Also fix the endptr condition which was backwards and resulted in every tweet id getting rejected, but you didn't see that one, this commit really is about the tweet id 00 which is the most important tweet id.

----

Had this one in a stash but forgot to commit/PR. Already tested and it works. Not much to review here.


nattofriends:
I was wondering why reply-by-id wasn't working recently...

dequis:
I don't know what "patience" is, and there's not much to review here, merging to develop right away.

PR #27, opened 2015-03-11
chat_name_hint: normalize utf8 with fallback from the channel name

seirl:
When using chat_name_hint(), normalize utf8 names with a fallback to ascii (it was already done that way for nicks.)
Example: #Voilà → #Voila

Tested with valgrind.
We could also add utf8 support for channels if utf8_nicks is enabled, but this change would be more complex.

### DIFF
--- a/irc_im.c
+++ b/irc_im.c
@@ -694,7 +694,8 @@ static gboolean bee_irc_chat_name_hint(bee_t *bee, struct groupchat *c, const ch
 {
 	irc_t *irc = bee->ui_data;
 	irc_channel_t *ic = c->ui_data, *oic;
-	char stripped[MAX_NICK_LENGTH + 1], *full_name;
+	char *stripped, *full_name;
+	gsize bytes_written;
 
 	if (ic == NULL) {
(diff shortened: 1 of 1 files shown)


Wilm0r:
BTW I think at the time I intentionally did not do this because the RFCs did not specify limitations on what is allowed in channel names, other than some obvious ASCII restrictions like no spaces.

seirl:
Currently the channel names are still stripped, see the irc_channel_name_strip() function that removes all the characters > 0x7f. This patch is better than just stripping everything. I agree that we should allow channels with utf8 characters but it should be a choice we leave to the user (with the utf8_nicks setting for instance), because a lot of clients/scripts/... just fail with utf8 channel names.

dequis:
Patch looks okay. Didn't test.

Relevant issue: http://bugs.bitlbee.org/bitlbee/ticket/1152

dequis:
Applied as b95d03b74c1186a67f0a5e5ddf909b496505a3b5 with two minor style changes.

PR #31, opened 2015-04-05
jabber: Register "hipchat" protocol (only minimal support for now)

dequis:
Another take on [the subprotocols idea][1] that, IMO, was a failure.

Unlike the other implementation, this one doesn't touch gtalk/facebook accounts, it just adds another copy of the "jabber" prpl called "hipchat".

And, based on the protocol name:

- sets JFLAG_HIPCHAT to jabber_data
- sets the default value of the "server" setting
- only includes the oauth setting for jabber-type accounts

This is slightly more "hardcoded" but honestly facebook and gtalk are just as hardcoded as this.

Copying the prpl is needed because the meaning of the usernames is completely different (there's no srv lookup stuff either)

[1]: https://github.com/bitlbee/bitlbee/commit/7733b8c95abbac26f5cbf02a9874a9d96c23bf0d

### DIFF
--- a/protocols/jabber/jabber.c
+++ b/protocols/jabber/jabber.c
@@ -60,8 +60,6 @@ static void jabber_init(account_t *acc)
 
 	s = set_add(&acc->set, "activity_timeout", "600", set_eval_int, acc);
 
-	s = set_add(&acc->set, "oauth", "false", set_eval_oauth, acc);
-
 	s = set_add(&acc->set, "display_name", NULL, NULL, acc);
 
 	g_snprintf(str, sizeof(str), "%d", jabber_port_list[0]);
@@ -82,6 +80,12 @@ static void jabber_init(account_t *acc)
 
 	s = set_add(&acc->set, "server", NULL, set_eval_account, acc);
 	s->flags |= SET_NOSAVE | ACC_SET_OFFLINE_ONLY | SET_NULL_OK;
+
+	if (strcmp(acc->prpl->name, "hipchat") == 0) {
+		set_setstr(&acc->set, "server", "chat.hipchat.com");
+	} else {
+		s = set_add(&acc->set, "oauth", "false", set_eval_oauth, acc);
+	}
 
 	s = set_add(&acc->set, "ssl", "false", set_eval_bool, acc);
 	s->flags |= ACC_SET_OFFLINE_ONLY;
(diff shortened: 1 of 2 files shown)


binwiederhier:
@dequis FTW! Good job!

protocols/jabber/jabber.c
```diff
@@ -83,6 +81,12 @@ static void jabber_init(account_t *acc)
 	s = set_add(&acc->set, "server", NULL, set_eval_account, acc);
 	s->flags |= SET_NOSAVE | ACC_SET_OFFLINE_ONLY | SET_NULL_OK;
 
+	if (strcmp(acc->prpl->name, "hipchat") == 0) {
+		set_setstr(&acc->set, "server", "chat.hipchat.com");
```
Wilm0r: Why this instead of passing a default to set_add()?

protocols/jabber/jabber.c
```diff
@@ -685,4 +694,9 @@ void jabber_initmodule()
 	ret->buddy_action = jabber_buddy_action;
 
 	register_protocol(ret);
+
+	/* Another one for hipchat, which has completely different logins */
```
Wilm0r: The only thing you're changing is the name property though?

But sure, copying a prpl is the way to create a new protocol.

Wilm0r:
> This is slightly more "hardcoded" but honestly facebook and gtalk are just as hardcoded as this.

They are indeed! I'm kinda ok with it really, don't mind it much here either.

The risk of actually defining a different protocol is it's a pain to revert to an old version with no support for the protocol. This is going to possibly get more painful when playing with the RPC stuff as well so I'm considering adding support for "dead" protocols - allow loading and re-saving the data as-is but ignore the account otherwise. So at least configs can be loaded and don't get corrupted on save.

protocols/jabber/jabber.c
dequis: Yup!

protocols/jabber/jabber.c
dequis: Because of @binwiederhier's bug, fixed in a51220df24aa87b638ef0064e3078ae8da644b4f, which says "so set_eval_account gets called" in the commit log.

Turns out it *looked* like it was set to the right value when it was a default, but without set_eval_account, it kept trying to do a SRV (in my case, it tried to connect to a gtalk server, heh).

It only applied to newly created accounts, too, so wouldn't have noticed otherwise. Sneaky bug.

dequis:
Yeah reverting this would be ugly. At least now it provides a slightly better error message when a protocol isn't found! It probably still removes everything that follows afterwards in the xml file though.

dequis:
beep

Wilm0r:
O hai. So, my only objection is troubles with rollbacks but that's not specific to this patch at all and you're not making that problem worse for GTalk/FB (of which at least FB is going away soon anyway).

So go for it.

PR #32, opened 2015-04-07
doc/Makefile: More reliable check for git/bzr tree

qsuscs:
Signed-off-by: Thomas Schneider <thosch97@gmail.com>

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -4,8 +4,8 @@ _SRCDIR_ := $(_SRCDIR_)doc/
 endif
 
 all:
-	# Only build the docs if this is a git tree
-	test ! '(' -d ../.git -o -d ../.bzr ')' || $(MAKE) -C user-guide
+	# Only build the docs if this is a git or bzr tree
+	( git rev-parse --is-inside-work-tree || bzr status ) && $(MAKE) -C user-guide
 
 install:
 	mkdir -p $(DESTDIR)$(MANDIR)/man8/ $(DESTDIR)$(MANDIR)/man5/


dequis:
You broke it. Not sure how, but you did

qsuscs:
I have no idea why Travis is failing.  I did not even change the file where it breaks and I cannot see any error in this file.  Try it yourself …

dequis:
>17:06 < dx> i think the doc/ directory of the debian package build has no Makefile
>17:06 < qsuscs> no, it has
>17:06 < qsuscs> line 886, make[3]: *** No rule to make target `user-guide.txt', needed by `all'.  Stop.
>17:06 < dx> i'm talking about debian/build-native/doc
>17:07 < qsuscs> so it is in doc/user-guide
>17:07 < qsuscs> aaand it has enough of a makefile to know it has to build user-guide.txt, it just does not know how
>17:07 < dx> \o/
>17:08 < qsuscs> to debug, i would probably throw in a `cat Makefile`, but as i said, i don’t have a debian-ish system here
>17:09 < dx> okay, new guess: Makefile is the only thing it has
>17:09 < dx> https://github.com/bitlbee/bitlbee/blob/master/configure#L174
>17:11 < dx> also the build-native and build-libpurple steps shouldn't be building the docs every time
>17:12 < qsuscs> err wtf
>17:12 < qsuscs> okay but it’s not me who broke this.
>17:12 < dx> there's a make -C doc/user-guide below in debian/rules for that
>17:12 < dx> nope, you did
>17:12 < dx> the -d ../git check fails when inside build-native/doc/
>17:12 < dx> but the check you added goes deeper than that
>17:12 < dx> and decides to build it anyway
>17:13 < dx> \o/
>17:13 < dx> sorry qsuscs

PR #33, opened 2015-04-21
hipchat: Basic implementation: Auth, profile and mention names

dequis:
This is enough to log in with their usernames, make 'chat add' based groupchat joins slightly more smooth, and see mention names as nicks.

All the MUC list stuff is left out intentionally since that's not as stable as I wish.

-----------

I said I was going to just push this thing, but it turned out to be short enough

The next PR is probably going to be 59c1fe7fce52811f6cd99345373bee26a478c27b which is mostly a major rewrite of jabber_chat_pkt_message that benefits normal jabber MUCs too

### DIFF
--- a/protocols/jabber/hipchat.c
+++ b/protocols/jabber/hipchat.c
@@ -1,0 +1,93 @@
+/***************************************************************************\
+*                                                                           *
+*  BitlBee - An IRC to IM gateway                                           *
+*  Jabber module - HipChat specific functions                               *
+*                                                                           *
+*  Copyright 2015 Xamarin Inc                                               *
+*                                                                           *
+*  This program is free software; you can redistribute it and/or modify     *
+*  it under the terms of the GNU General Public License as published by     *
+*  the Free Software Foundation; either version 2 of the License, or        *
+*  (at your option) any later version.                                      *
+*                                                                           *
+*  This program is distributed in the hope that it will be useful,          *
+*  but WITHOUT ANY WARRANTY; without even the implied warranty of           *
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the            *
+*  GNU General Public License for more details.                             *
+*                                                                           *
+*  You should have received a copy of the GNU General Public License along  *
+*  with this program; if not, write to the Free Software Foundation, Inc.,  *
+*  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.              *
+*                                                                           *
+\***************************************************************************/
+
+#include "jabber.h"
+
+xt_status hipchat_handle_success(struct im_connection *ic, struct xt_node *node)
+{
+	struct jabber_data *jd = ic->proto_data;
+	char *sep, *jid;
+
+	jid = xt_find_attr(node, "jid");
+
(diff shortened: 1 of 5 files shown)


protocols/jabber/sasl.c
```diff
@@ -142,15 +147,33 @@ xt_status sasl_pkt_mechanisms(struct xt_node *node, gpointer data)
...
 		/* With SASL PLAIN in XMPP, the text should be b64(\0user\0pass) */
-		len = strlen(jd->username) + strlen(ic->acc->pass) + 2;
-		s = g_malloc(len + 1);
-		s[0] = 0;
-		strcpy(s + 1, jd->username);
-		strcpy(s + 2 + strlen(jd->username), ic->acc->pass);
+		g_string_append_c(gs, '\0');
+		g_string_append(gs, username);
+		g_string_append_c(gs, '\0');
+		g_string_append(gs, ic->acc->pass);
```
jgeboski: `g_string_append_printf()`?

protocols/jabber/hipchat.c
```diff
@@ -0,0 +1,97 @@
...
+	/* Hipchat's auth doesn't expect a restart here */
+	jd->flags &= ~JFLAG_STREAM_RESTART;
+
+	if (!jabber_get_roster(ic)) {
+		return XT_ABORT;
+	}
+	if (!jabber_iq_disco_server(ic)) {
+		return XT_ABORT;
+	}
+	if (!jabber_get_hipchat_profile(ic)) {
```
jgeboski: Holy `if`s!

protocols/jabber/sasl.c
Wilm0r: I assume the \0s could cause troubles. Maybe %c (and then 0 as an argument) would work but it's still ugly.

dequis: hmmm, drafting it here

```
g_string_append_printf(gs, "%c%s%c%s", 0, username, 0, ic->acc->pass);

if (want_hipchat) {
    g_string_append_printf(gs, "%c%s", 0, set_getstr(&ic->acc->set, "resource"))
}
```

I dunno.


How about

```
char *pass = ic->acc->pass;
char *resource = set_getstr(&ic->acc->set, "resource");
if (want_hipchat) {
    g_string_printf(gs, "%c%s%c%s%c%s", 0, username, 0, pass, 0, resource);
} else {
    g_string_printf(gs, "%c%s%c%s", 0, username, 0, pass);
}
```

Maybe better? (I don't know if it works, though.)

Wilm0r: Meh, There are many ways to overengineer this. It's XMPP authentication which even on a shared machine you probably do once a minute or so at most so why optimise it.

If you could do "\0%\0%s..." that'd be nice, but not possible, so I'd just keep what you already have.

jgeboski: Yeah, it was a suggestion without much thought.  I do not think there is anyway guaranteed way for `sprintf()` to handle null-terminating chars.  Disregard my question!

dequis:
Can I consider this reviewed?

jgeboski:
I would say push it into `develop` for more testing.

PR #34, opened 2015-05-02
Twitter "url" command and retweet-with-comment support

Wilm0r:
Let me try this pull request thing. If you feel like pointing out anything retarded please do, while I spend the next two days trying to figure out how to do merges with Git.

### DIFF
--- a/doc/user-guide/commands.xml
+++ b/doc/user-guide/commands.xml
@@ -120,7 +120,7 @@
 
 			<description>
 				<para>
-					This commands deletes an account from your account list. You should signoff the account before deleting it.
+					This command deletes an account from your account list. You should signoff the account before deleting it.
 				</para>
 
 
@@ -852,6 +852,7 @@
 				<varlistentry><term>unfollow &lt;screenname&gt;</term><listitem><para>Stop following a person</para></listitem></varlistentry>
 				<varlistentry><term>favourite &lt;screenname|#id&gt;</term><listitem><para>Favo<emphasis>u</emphasis>rite the given user's most recent tweet, or the given tweet ID.</para></listitem></varlistentry>
 				<varlistentry><term>post &lt;message&gt;</term><listitem><para>Post a tweet</para></listitem></varlistentry>
+				<varlistentry><term>url &lt;screenname|#id&gt;</term><listitem><para>Show URL for a tweet to open it in a browser (and see context)</para></listitem></varlistentry>
 			</variablelist>
 
 			<para>
--- a/protocols/twitter/twitter.c
+++ b/protocols/twitter/twitter.c
@@ -1002,6 +1002,19 @@ static void twitter_handle_command(struct im_connection *ic, char *message)
 		message = cmd[2];
 		in_reply_to = id;
 		allow_post = TRUE;
+	} else if (g_strcasecmp(cmd[0], "url") == 0) {
+		id = twitter_message_id_from_command_arg(ic, cmd[1], &bu);
+		if (!id) {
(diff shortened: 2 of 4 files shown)


dequis:
Noice.

Looks good... except for a detail that everyone gets wrong and i'm not sure if i want to keep it like that: I decided that pull requests should be submitted against the 'develop' branch instead of 'master', but since master is still the default branch, pull requests go there by default unless you change it manually. And this results in wasting time when someone gets it wrong, because the only way to fix it is to close and reopen \o/.

Not that anything important is lost, especially in this case where both branches point to the same commit. It would be mostly metadata, such as the difference between a PR that gets 'closed' vs 'merged'.

>Also, fix bug in parsing of entities in DMs.

Needs more `git add -p` to add that part as a different commit (but meh, it's okay)

Wilm0r:
Ah, I used master as a base intentionally, and will do the merge by hand instead of with the magic button here. So then the PR target doesn't really matter.

> Needs more git add -p to add that part as a different commit (but meh, it's okay)

That'd be nice yeah. Though in this case this more or less just happened as I was updating the code to the new invocation of expand_entities(). Silly enough the DM parser had an entities variable which was getting passed to expand_entities() ... and no code to set that var to anything other than NULL. D'oh.

dequis:
> Ah, I used master as a base intentionally

I don't really follow the logic buuuut

>and will do the merge by hand instead of with the magic button here. So then the PR target doesn't really matter

...yeah, it never really matters when i do it either since I always merge stuff that way, my main concern is metadata, which isn't a particularly real concern.

------

Regarding the code: I kinda had a hard time following that weird %s%s%s%s%s%s stuff you did.

Can we throw newlines in there?

Sorta like

>00:19:03 < BitlBeeTest2> [13] ok nog eentje <https://t.co/eQaBK01Xvk>
>00:19:03 < BitlBeeTest2> RT @ Wilmer: After some hours on the UK roads,I don't see why people fear self-driving cars. Can't be hard to have computers drive at least that "well".

vs

>00:19:03 < BitlBeeTest2> [13] ok nog eentje <https://t.co/eQaBK01Xvk> [@ Wilmer: After some hours on the UK roads,I don't see why people fear self-driving cars. Can't be hard to have computers drive at least that "well".]

That would mean reordering the format strings

Old: `%(text)s <%(kort)s> [%(disp)s]%(rest)s`
New: `%(text)s %(kort)s %(rest)s\nRT %(quoted_user)s: %(quoted_text)s`

(using imaginary python-style format strings for the example)

Although, hm, it might become problematic with the strip_newlines setting. So maybe no, we can't really throw newlines in there.

I dunno. Maybe just merge this, much better than nothing for sure

Wilm0r:
We should at some point have a chat about the existence of master and develop and how you see their purposes - My plan is to put most stuff into master and riskier stuff into develop first. And I don't do risky stuff. :>

Format strings: So is it about readability of the string? Could do something like this:

```C
format = "%s" // current status up to URL
         "<%s>" // t.co URL to quotedstatus
         "[%s]" // original status in brackets
         "%s" // rest of current/quoting status (should be empty)
```

Readability in the UI: Many people hate newlines in Tweets, and the example above looks a little ugly to me as well. :-/ I'd just keep this format.

dequis:
> ```C
format = "%s" // current status up to URL
         "<%s>" // t.co URL to quotedstatus
         "[%s]" // original status in brackets
         "%s" // rest of current/quoting status (should be empty)
```

Ehhhh... nah. Whatever. It's fine.

The last commit, however, is not fine.

Wilm0r:
Yeah, that's definitely not one of my finest commits ever. If you have a better way let me know.

PR #36, opened 2015-05-16
Add python script to generate help.txt, replacing xslt stuff

dequis:
This would mean changing one build dependency for another, but one that is way more common at least. (Fun fact: the xslt stuff depends on perl)

It generates *almost* the same thing as the xslt - a bit better if you ask me, since it correctly handles a few <emphasis> tags in the middle of the text, which were previously stripped. One example of that is:

    Favo<emphasis>u</emphasis>rite the given user [...]

Outputs `Favo\x02u\x02rite` with this script, `Favorite` with the xslt. (That's actually an accidental feature)

The script works in python2 and python3 and only uses the stdlib

-----

The full diff between the old generated help.txt and the new help.txt

```diff
--- help.txt	2015-05-16 01:52:21.000000000 -0300
+++ help2.txt	2015-05-16 01:46:06.000000000 -0300
@@ -57,7 +57,7 @@
 Now you might want to add some contacts, to do this we will use the \x02add\x02 command. It needs two arguments: a connection ID (which can be a number (try \x02account list\x02), protocol name or (part of) the screenname) and the user's handle. It is used in the following way: \x02add <connection> <handle>\x02
 
 \x02<you>\x02 add 0 r2d2@example.com
-\x02* r2d2\x02  has joined &bitlbee
+\x02* r2d2\x02  has joined \x02&bitlbee\x02
 
 In this case r2d2 is online, since he/she joins the channel immediately. If the user is not online you will not see them join until they log on.
 
@@ -276,7 +276,7 @@
 
 \x02Example:\x02
 \x02<ctrlsoft>\x02 add 3 gryp@jabber.org grijp
-\x02* grijp\x02 has joined &bitlbee
+\x02* grijp\x02 has joined \x02&bitlbee\x02
 %
 ?info
 \x02Syntax:\x02 info <connection> <handle>
@@ -286,7 +286,7 @@
 
 \x02Example:\x02
 \x02<ctrlsoft>\x02 info 0 72696705
-\x02<root>\x02 User info - UIN: 72696705   Nick: Lintux   First/Last name: Wilmer van der Gaast   E-mail: lintux@lintux.cx
+\x02<root>\x02 User info - UIN: 72696705 Nick: Lintux First/Last name: Wilmer van der Gaast E-mail: lintux@lintux.cx
 %
 ?remove
 \x02Syntax:\x02 remove <nick>
@@ -295,7 +295,7 @@
 
 \x02Example:\x02
 \x02<ctrlsoft>\x02 remove gryp
-\x02* gryp\x02 has quit [Leaving...]
+\x02* gryp\x02 has quit \x02[Leaving...]\x02
 %
 ?block
 \x02Syntax:\x02 block <nick>
@@ -426,7 +426,7 @@
 
 \x02Example:\x02
 \x02<itsme>\x02 rename itsme_ you
-\x02* itsme_\x02 is now known as you
+\x02* itsme_\x02 is now known as \x02you\x02
 %
 ?yes
 \x02Syntax:\x02 yes [<number>]
@@ -661,7 +661,7 @@
  \x02report <screenname|#id>\x02 - Report the given user (or the user who posted the tweet with the given ID) for sending spam. This will also block them.
  \x02follow <screenname>\x02 - Start following a person
  \x02unfollow <screenname>\x02 - Stop following a person
- \x02favourite <screenname|#id>\x02 - Favourite the given user's most recent tweet, or the given tweet ID.
+ \x02favourite <screenname|#id>\x02 - Favo\x02u\x02rite the given user's most recent tweet, or the given tweet ID.
  \x02post <message>\x02 - Post a tweet
  \x02url <screenname|#id>\x02 - Show URL for a tweet to open it in a browser (and see context)
 
```

### DIFF
--- a/configure
+++ b/configure
@@ -687,16 +687,6 @@ if [ "$doc" = "1" ]; then
 		exit 1
 	fi
 	echo "DOC=1" >> Makefile.settings
-
-	if [ "$skype" = "1" -o "$skype" = "plugin" ]; then
-		# skype also needs asciidoc
-		if ! type a2x > /dev/null 2> /dev/null; then
-			echo
-			echo 'WARNING: The skyped man page requires asciidoc. It will not be generated.'
-		else
-			echo "ASCIIDOC=1" >> Makefile.settings
-		fi
-	fi
 fi
 
 get_version
--- a/configure
+++ b/configure
@@ -680,16 +680,13 @@ if [ "$skype" = "1" -o "$skype" = "plugin" ]; then
 fi
 
 if [ "$doc" = "1" ]; then
-	if [ ! -e doc/user-guide/help.txt ] && \
-	     ! type xmlto > /dev/null 2> /dev/null || \
-	     ! type xsltproc > /dev/null 2> /dev/null
-	then
+	# check this here just in case someone tries to install it in python2.4...
+	if ! python -m xml.etree.ElementTree > /dev/null 2>&1; then
 		echo
-		echo 'WARNING: Building from an unreleased source tree without prebuilt helpfile.'
-		echo 'Install xmlto and xsltproc if you want online help to work.'
-	else
-		echo "DOC=1" >> Makefile.settings
+		echo 'ERROR: Python (>=2.5 or 3.x) is required to generate docs'
+		exit 1
 	fi
+	echo "DOC=1" >> Makefile.settings
 
 	if [ "$skype" = "1" -o "$skype" = "plugin" ]; then
 		# skype also needs asciidoc
--- a/doc/user-guide/genhelp.py
+++ b/doc/user-guide/genhelp.py
@@ -1,5 +1,8 @@
 #!/usr/bin/env python
-#
+
+# Usage: python genhelp.py input.xml output.txt
+# (Both python2 (>=2.5) or python3 work)
+
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
 # as published by the Free Software Foundation; either version 2
@@ -17,10 +20,9 @@
 
 
 import re
+import sys
 import xml.etree.ElementTree as ET
 
-IN_FILE = 'help.xml'
-OUT_FILE = 'help.txt'
 NORMALIZE_RE = re.compile(r"([^<>\s\t])[\s\t]+([^<>\s\t])")
 
 # Helpers
@@ -215,8 +217,12 @@ def tag_simplelist(tag, parent):
 
 
 def main():
-    txt = process_file(IN_FILE)
-    open(OUT_FILE, "w").write(txt)
+    if len(sys.argv) != 3:
+        print("Usage: python genhelp.py input.xml output.txt")
+        return
+
+    txt = process_file(sys.argv[1])
+    open(sys.argv[2], "w").write(txt)
 
 if __name__ == '__main__':
     main()
--- a/doc/user-guide/genhelp.py
+++ b/doc/user-guide/genhelp.py
@@ -1,0 +1,165 @@
+import re
+import xml.etree.ElementTree as ET
+
+IN_FILE = 'help.xml'
+OUT_FILE = 'help.txt'
+NORMALIZE_RE = re.compile(r"([^<>\s\t])[\s\t]+([^<>\s\t])")
+
+def join(list):
+    return ''.join([str(x) for x in list])
+
+def normalize(x):
+    x = NORMALIZE_RE.sub(r"\1 \2", x or '')
+    return x.replace("\n", "").replace("\t", "")
+
+def fix_tree(tag, lvl=''):
+    if tag.tag.count("XInclude"):
+        tag.tag = 'include'
+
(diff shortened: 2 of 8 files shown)


doc/user-guide/genhelp.py
```diff
@@ -0,0 +1,165 @@
+import re
```
Wilm0r: No hashbang?
Also, zomg spaces? :-P (But sure, if you really really like that, just keep it out of the C files pls!)

doc/user-guide/genhelp.py
```diff
@@ -0,0 +1,165 @@
...
+
+    for subtag in tag:
+        fix_tree(subtag, lvl + "  ")
+
+    #print("%s</%s>%r" % (lvl, tag.tag, [tag.tail, normalize(tag.tail)]))
+
+    tag.text = normalize(tag.text)
+    tag.tail = normalize(tag.tail)
+
+def parse_file(filename, parent=None):
```
Wilm0r: It parses and processes, it does everything! \o/

Wilm0r:
Gee thanks for totally showing the exact content I cared about shithub!

Not the prettiest and clearest-commented code I've seen. :-P But I don't care much since this is just the doc generator. The output looks fine.

This takes help.xml, ET takes care of the includes already so stuff just works?

I asssume the first parse_tag turns into parse_book and everything else happens from there?

doc/user-guide/genhelp.py
dequis: Spaces are what PEP 8 recommends https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces - and they just feel right for me when dealing with python code. Only happens with python, oddly enough.

Of course I won't use them for the C files.

Hashbang, welp.

dequis:
>Not the prettiest and clearest-commented code I've seen. :-P

Prettier than help.xsl! :3333

But yeah I don't believe in comments. Or docstrings. I probably should.

This actually started with an ipython notebook that looked *way* uglier than this and I only started deuglifying it after it generated sane output. I kept the `globals()["tag_%s" % ...]` thing because it's really handy, though.

>This takes help.xml, ET takes care of the includes already so stuff just works?

Yeah, `tag_include` calls `parse_file`

>I asssume the first parse_tag turns into parse_book and everything else happens from there?

Well, turns into `tag_book`, but yeah pretty much that (and `tag_book` is just aliased to `parse_subtags`)

doc/user-guide/genhelp.py
dequis: Hashbang added up there.

dequis:
Threw three commits there, now actually replacing all the xmlto/xsltproc stuff with this script.

Something I'm not *completely* sure about:

>Build only help.txt (with genhelp.py), create 'user-guide' target
>
>Since user-guide.{txt,html} aren't normally needed for anything and add xsltproc/xmlto as dependencies.

...are user-guide.{txt,html} actually needed for anything i'm not aware of?

OH NEVERMIND travis just answered that for me as the build status for the last commit just went from 'pending' to 'failed'. Yep, the debian doc package.

configure
```diff
@@ -680,16 +680,13 @@ if [ "$skype" = "1" -o "$skype" = "plugin" ]; then
...
-	     ! type xsltproc > /dev/null 2> /dev/null
-	then
+	# check this here just in case someone tries to install it in python2.4...
+	if ! python -m xml.etree.ElementTree > /dev/null 2>&1; then
 		echo
-		echo 'WARNING: Building from an unreleased source tree without prebuilt helpfile.'
-		echo 'Install xmlto and xsltproc if you want online help to work.'
-	else
-		echo "DOC=1" >> Makefile.settings
+		echo 'ERROR: Python (>=2.5 or 3.x) is required to generate docs'
```
Wilm0r: And I think elementtree is included by default these days right? (Chcked: Yes, it is.)

Wilm0r:
So as you say there are still issues with user-guide, no?

dequis:
I posted that comment right after I pushed 965127b22c80fef806fdbfa84c52e0671a2f0826 which fixed those issues.

Also I just checked debian build-deps and...

>Build-Depends: libglib2.0-dev (>= 2.4), libevent-dev, libgnutls28-dev | libgnutls-dev | gnutls-dev, po-debconf, libpurple-dev, libotr5-dev, debhelper (>= 6.0.7~), asciidoc

xmlto/xsltproc weren't there because they are covered by asciidoc, which is still (sorta) a dependency because of the skyped manpage.

In fact, the last commit that removes the dependencies from .travis.yml, 8cbe2b6d2b82fa071e136ffae3a7873e969f3dd2, is mostly a no-op because it's still installing asciidoc. Welp.

I'll happily murder skyped when the msnp24 stuff is done.

dequis:
After posting that comment I went "fuck it, no one is going to change that manpage ever again anyway" and added `skyped.1` to the repo in 09bd226ff479c3ab3b265c1ae721c3ca190d3c83, additionally removing all the remaining references to asciidoc.

:D

So yeah, I think now i *really* am done here.

dequis:
beep

PR #38, opened 2015-05-26
Lowering xmpp presence priority on away.

sm00th:
When user set's away lower xmpp presence priority by 5 as most clients do, new
priority won't go below zero though.

Basically the same patch as at https://bugs.bitlbee.org/bitlbee/ticket/1164 but with updated codestyle.
Not sure if this needs a separate setting to choose the away priority.

### DIFF
--- a/protocols/jabber/presence.c
+++ b/protocols/jabber/presence.c
@@ -180,6 +180,22 @@ xt_status jabber_pkt_presence(struct xt_node *node, gpointer data)
 	return XT_HANDLED;
 }
 
+static char *choose_priority(struct im_connection *ic)
+{
+	struct jabber_data *jd = ic->proto_data;
+	char *prio = set_getstr(&ic->acc->set, "priority");
+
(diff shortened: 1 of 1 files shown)


protocols/jabber/presence.c
```diff
@@ -180,6 +180,21 @@ xt_status jabber_pkt_presence(struct xt_node *node, gpointer data)
...
 }
 
+static char *choose_priority(struct im_connection *ic)
+{
+	struct jabber_data *jd = ic->proto_data;
+	char *prio = set_getstr(&ic->acc->set, "priority");
+
+	if (jd->away_state->code != NULL) {
+		int new_prio = (atoi(prio) - 5);
+		if (new_prio < 0)
```
dequis: style nitpick: {...} are required now even for one line blocks

dequis:
Merged to develop as 24f113b01e1117ab031a1ede6d418030fc69d63d. Thanks!

PR #39, opened 2015-05-28
Fixes 'Provides' in debian control file

zer0def:


debian/control
```diff
@@ -21,7 +21,7 @@ Description: An IRC to other chat networks gateway (default version)
 Package: bitlbee-libpurple
 Architecture: any
 Depends: ${misc:Depends}, ${shlibs:Depends}, debianutils (>= 1.16), bitlbee-common (= ${source:Version})
-Provides: bitlbee
+Provides: bitlbee (= ${source:Version})
```
Wilm0r: Source and binary version are not always going to be the same IIRC (binary NMUs etc). I think there's another variable you can use for the currently-being-built binary version. But don't remember what it was, let's see first how this works I guess.

PR #41, opened 2015-06-04
Revert "Provides" "fix" in debian control file.

zer0def:
As "Provides" doesn't seem to work as expected, if plugin makers are sure their code works with -libpurple, they have to explicitly set an alternate dependency.


dequis:
"Revert" "Provides" "fix" "in" "debian" "control" "file"

PR #43, opened 2015-08-03
debian packaging updates

wrayan:
HI,

applied some minor changes to the debian tree while packaging 3.4.1 for im.kernel-oops.de. Maybe you want to adopt some of them.
- added upstart script
- removed debian from .gitignore
- removed obsolete source/options file (we're not on bzr any more
  and git-buildpackage will automatically run dpkg-source with -i)
- changed patches to quilt style and bumped source format to 3.0 (quilt)
- bumped standards version to 3.9.6 to satisfy lintian; no other
  changes


.gitignore
```diff
@@ -4,7 +4,6 @@ bitlbee
 user-guide.txt
 user-guide.html
 help.txt
-debian
```
dequis: This was intentional, so that only explicitly added files are tracked by git (and excluding all the automatically generated stuff). How does it improve things for you?

Wilm0r:
> removed obsolete source/options file

I've done a s/bzr/git/ on that one yesterday for the 3.4.1-1 upload. Why assume everyone's using a git-specific package build wrapper if you can just keep things working nicely?

.gitignore
wrayan: Hmm ok that's probably a matter of taste ;) I usually don't expect there to be any auto-generated stuff left after building with gbp or debuild so if git status shows me any untracked files in debian/ I know either something is broken or I forgot to add something I changed - if the dir is ignored, I could miss something important.

wrayan:
@Wilm0r hum ok I missed that. You're right, your solution is better.

dequis:
So, uh, are you planning on changing stuff (like fixing the merge conflict) or should i just go ahead and pick parts of this thing here and there?

wrayan:
I'm sorry, the conflict should be resolved now and I reverted my changes to .gitignore and source/options.

The ci test is failing due to missing orig tarball though. There are several options to resolve this but I'm not sure what you prefer. As debian packaging stuff is done by upstream, instead of 3.0 (quilt) the format could be switched to 3.0 (native). In that case however, there should be no distro-specific changes - the original source is expected to be distro native thus including all distro-specific stuff already, and the package version should also not contain a debian revision number. I'd definately prefer 3.0 (quilt) to keep packaging stuff separate. Of course you can also just stay with the old 1.0 format if you don't care about the shiny things mentioned here https://wiki.debian.org/Projects/DebSrc3.0 :)

dequis:
Hmmm, I don't know a lot about this stuff, when i did [the debian crap for bitlbee-facebook][1] i didn't include a source format - does that mean it's `3.0 (native)`? In that case, everything worked just fine with a directory structure that is very similar to this one. No patches, though.

[1]: https://github.com/jgeboski/bitlbee-facebook/commit/ea5d6af3cecd01222172324d320718ff629e3a7f

wrayan:
Nope it will fallback to 1.0 then. I took a look at the bitlbee-facebook repo and tried to build a source package from it. Here's what happened: http://pastebin.com/8sZkMZiW

Your version string consists of <upstream_version>-<debian-revision> - this indicates that the source used for the debian package differs from upstream and/or packaging is separately maintained.

As mentioned earlier, usually a non-native package would consist of the "vanilla" tarball and a single diff (1.0 format) or a tarball (3.0 format) containing the debian specific stuff whereas a native source package would consist of only one tarball containing all the files including the debian/ tree. Both are fine but if you want this to be a native package, you should remove the revision number from the version string in changelog because there is no diff between vanilla and debian source that would require separate versioning.

As mentioned earlier, usually a non-native package would consist of the vanilla tarball and a single diff (1.0 format) or a tarball (3.0 format) containing the debian specific stuff whereas a native source package would consist of only one tarball containing all the files including the debian/ tree. Both are fine but if you want this to be a native package, you should remove the revision number from the version string in changelog because there is no separately maintained stuff that would require separate versioning.

Native packages are also not supposed to contain patches as there is no need for them if your source is written solely for debian. Of course you could still include patches if you apply them manually in your rules file but that's not the best of all ideas.

So, yes, you're ending up with a technically functional source package but from a package maintainer's perspective it's still bad. You don't need to care about that if you only want to package it for yourself, it won't break your system or eat your cat, but you could not upload it to debian or a launchpad ppa for example, it would be rejected.

I was struggling with this for quite some time, too and finally ended up in always using quilt format. Native packages really only make sense if you don't care about users of other distros at all or the code is of no use for them anyway. I set up my git repo to have the debian stuff in a separate branch so I just need to check out the "vanilla" branch to create that orig tarball. There are even helper tools around to do that (among some fancy patch management stuff) automatically but I suppose that's only of interest for people who are packaging a lot.

If you want to dive deeper into the subject, I suggest we continue discussion on irc. Though debian packaging is quite logical it can be quite complex and difficult to explain ;)

jspricke:
+1 for fixing and merging this (except the upstart job maybe ;), @wrayan ping me, if you need help with it).
Regarding the `source/format`, I would propose to go with `3.0 (quilt)` as well, as it's the current version and native would mean we can have only one Debian package version per upstream version ;).
Also, I would proose to merge the patched bitlbee.conf upstream ;).

debian/control
```diff
@@ -3,7 +3,7 @@ Section: net
 Priority: optional
 Maintainer: Wilmer van der Gaast <wilmer@gaast.net>
 Uploaders: Jelmer Vernooij <jelmer@debian.org>
-Standards-Version: 3.9.5
+Standards-Version: 3.9.6
```
jelmer: This should be 3.9.8 now, as two more revisions have come out since this MP was submitted.

debian/patches/0001-bitlbee.conf.patch
```diff
@@ -1,5 +1,12 @@
---- bitlbee.conf	2009-06-01 00:20:24.000000000 +0100
-+++ bitlbee.conf	2009-06-07 21:16:19.000000000 +0100
+
```
jelmer: Why this change? The current patch file could use improvements e.g. by adding DEP-3 tags, but changing from one arbitrary patch format (AFAICT) to another seems unnecessary.

dequis:
Honestly I should have closed this one, it's stale and smelling bad. Better open a fresh PR if you want this in.

I still don't see the point of changing the source format, it breaks things (see travis errors) and the old format isn't going away any time soon. The only solutions i've been offered is to stop using dpkg-buildpackage or to stop using patches. I prefer staying with the current thing, which works.

>Also, I would proose to merge the patched bitlbee.conf upstream ;).

The diff is about debian specific changes. Also, we can't ship a bitlbee.conf that depends on the "bitlbee" system user existing, because that would break systems that don't have that user, or non-system-wide installations.

wrayan:
I agree with @dequis - this change is outdated and should be closed. I'd still vote for quilt source format though as it's the standard way for handling packaging-specific patches in deb sources.

jelmer:
@wilmer uses the packaging in master to build packages for older versions of Debian, some of which don't have support for source format 3 yet.

Or at least that was the reason we didn't switch to v3 earlier. Not sure if it still applies.

PR #44, opened 2015-08-07
typofix - https://github.com/vlajos/misspell_fixer

vlajos:


dequis:
Thanks!

(i actually used your tool before when i needed a filler commit for something, but that never got through. was a long time ago, before i got commit bit)

PR #46, opened 2015-09-11
IRCv3 SASL capability + PLAIN method

dequis:
Only plain, no other methods. We don't have built-in SSL to implement
EXTERNAL (certfp) and nothing else is worth implementing.

The actual authentication is pretty much like sending a server password
(when the server's authmode isn't closed), which means it happens in
cmd_identify, but this code also calls storage_check_pass() to send the
required success/failure replies.

SASL doesn't give us much benefit other than standards compliance, but
some clients might appreciate it.

And having a fourth way to do the same thing doesn't hurt! Now we have:

- identify in &bitlbee
- identify to nickserv (alias for root)
- server password
- sasl plain

--------

Pull request between two WIP branches because why not.

Spec: http://ircv3.net/specs/extensions/sasl-3.1.html

Wrote this along with the review of https://github.com/irssi/irssi/pull/278

### DIFF
--- a/irc_commands.c
+++ b/irc_commands.c
@@ -79,11 +79,11 @@ static gboolean irc_sasl_plain_parse(char *input, char **user, char **pass)
 	}
 
 	/* sanity checks */
-	if (part != 3 || i != (len + 1) || strcmp(parts[0], parts[1]) != 0) {
+	if (part != 3 || i != (len + 1) || (parts[0][0] && strcmp(parts[0], parts[1]) != 0)) {
 		g_free(decoded);
 		return FALSE;
 	} else {
-		*user = g_strdup(parts[0]);
+		*user = g_strdup(parts[1]);
 		*pass = g_strdup(parts[2]);
 		g_free(decoded);
 		return TRUE;
--- a/irc_commands.c
+++ b/irc_commands.c
@@ -27,6 +27,7 @@
 #include "bitlbee.h"
 #include "help.h"
 #include "ipc.h"
+#include "base64.h"
 
 static void irc_cmd_pass(irc_t *irc, char **cmd)
 {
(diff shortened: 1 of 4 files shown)


grawity:
> and nothing else is worth implementing.

Well SCRAM-SHA-1 or -256 may be useful sometime later, especially with said lack of native SSL.

dequis:
Hmmm, that made me think about how would the challenge/response thing work, and I guess we're completely stuck with PLAIN because we need the identify password to use it as the key to decrypt the IM account passwords from the storage. And SCRAM would not get us the plaintext key.

Even EXTERNAL, as I mentioned before, wouldn't work, unless the user's account is somehow configured to require certfp auth first, and only accept the identify password later through some other method. I guess someone could consider that an extra layer of security. I'm not sure if it's worth bothering.

(Native SSL isn't a huge priority since the most common way to use bitlbee is to connect to localhost, and some public servers have stunnel. Yeah, it's less than ideal.)

irc_commands.c
```diff
@@ -57,6 +58,117 @@ static void irc_cmd_pass(irc_t *irc, char **cmd)
...
+	int len;
+	int i = 0;
+	int part = 0;
+	guint8 *decoded;
+	char *parts[2];
+
+	/* bitlbee's base64_decode wrapper adds an extra null terminator at the end */
+	len = base64_decode(input, &decoded);
+
+	while (i < len && part < 3) {
```
Wilm0r: Seems odd to use while (with i initialised above) instead of just for?

irc_commands.c
```diff
@@ -57,6 +58,117 @@ static void irc_cmd_pass(irc_t *irc, char **cmd)
...
+		parts[part] = (char *) decoded + i;
+		i += strlen(parts[part]) + 1;
+		part++;
+	}
```
Wilm0r: Also ... wtf does this do? :-/

irc_commands.c
```diff
@@ -57,6 +58,117 @@ static void irc_cmd_pass(irc_t *irc, char **cmd)
...
+	if (status == STORAGE_OK) {
+		if (!irc->user->nick) {
+			/* set the nick here so we have it for the following numeric */
+			irc->user->nick = g_strdup(user);
+		}
+		irc_send_num(irc, 903, ":Password accepted");
+		return TRUE;
+
+	} else if (status == STORAGE_INVALID_PASSWORD) {
+		irc_send_num(irc, 904, ":Incorrect password");
```
Wilm0r: Are we providing that much detail elsewhere BTW? I thought BitlBee was paranoid and reported the same failure whether username or password was wrong.

irc_commands.c
dequis: Yeeeep probably should add a comment somewhere.

Sets parts[0] = start of first string, moves cursor forward to the next null with strlen, sets parts[1] to start of second string, etc.

irc_commands.c
dequis: I think `for (i = 0, part = 0; i < len && part < 3; part++)` could work. Maybe with some variable renaming

irc_commands.c
dequis: ag INVALID_PASSWORD -A2

```
root_commands.c
152:    case STORAGE_INVALID_PASSWORD:
153-            irc_rootmsg(irc, "Incorrect password");
154-            break;
263:    case STORAGE_INVALID_PASSWORD:
264-            irc_rootmsg(irc, "Password invalid");
265-            break;
```

First one is `cmd_identify`, the other is `cmd_drop`

irc_commands.c
```diff
@@ -57,6 +58,117 @@ static void irc_cmd_pass(irc_t *irc, char **cmd)
...
+			irc_send_num(irc, 902, ":Your SASL username does not match your nickname");
+
+		} else if (irc_sasl_check_pass(irc, user, pass)) {
+			/* and here we do the same thing as the PASS command*/
+			if (irc->status & USTATUS_LOGGED_IN) {
+				char *send_cmd[] = { "identify", pass, NULL };
+				root_command(irc, send_cmd);
+			} else {
+				/* no check_login here - wait for CAP END */
+				irc_setpass(irc, pass);
```
Wilm0r: Hrm, what's CAP END? I guess I should just read the spec, sorry. This means the SASL auth was done befure the USER and NICK cmd were sent IIRC, so if irc_setpass() set the password now, auto-login magic will happen later?

irc_commands.c
```diff
@@ -57,6 +58,117 @@ static void irc_cmd_pass(irc_t *irc, char **cmd)
...
+			}
+		}
+
+		g_free(user);
+		g_free(pass);
+
+	} else if (irc->status & USTATUS_IDENTIFIED) {
+		irc_send_num(irc, 907, ":You have already authenticated using SASL");
```
Wilm0r: USTATUS_IDENTIFIED is not SASL-specific BTW right?

root_commands.c
```diff
@@ -162,6 +162,12 @@ static void cmd_identify(irc_t *irc, char **cmd)
 		irc->status |= USTATUS_IDENTIFIED;
 		irc_umode_set(irc, "+R", 1);
 
+		if (irc->caps & CAP_SASL) {
+			irc_user_t *iu = irc->user;
+			irc_send_num(irc, 900, "%s!%s@%s %s :You are now logged in as %s",
+				iu->nick, iu->user, iu->host, iu->nick, iu->nick);
+		}
+
```
Wilm0r: Neat, so the SASL auth status *can* be set even if identifying through other means?

Should this maybe also be done after registering then?

irc_commands.c
dequis: Indeed. Could remove that part from the message. It's copypasted from the spec http://ircv3.net/specs/extensions/sasl-3.1.html

irc_commands.c
dequis: >The CAP END subcommand
>
>The END subcommand signals to the server that capability negotiation is complete and requests that the server continue with client registration. If the client is already registered, this command MUST be ignored by the server.
>
>Clients that support capabilities but do not wish to enter negotiation SHOULD send CAP END upon connection to the server. 

Basically it's a method to put the connection registration (the part of the connection that we call USTATUS_LOGGED_IN here, or everything before the 001 RPL_WELCOME numeric) on hold until the client declares that it's done.

And yes, your guess is correct. It's the same thing PASS does.

root_commands.c
dequis: >Should this maybe also be done after registering then?

Hmm, yep, I thought this covered all identification methods but there's another `irc->status |= USTATUS_IDENTIFIED;` in cmd_register(). Worth adding there too.

(BTW, I'm not sure how meaningful this 900 RPL_LOGGEDIN numeric is for clients, especially when it's not the result of AUTHENTICATE. It probably varies between implementations)

irc_commands.c
Wilm0r: No strong feelings. Mostly hated the initialisation being separate. (Think of the typical case of someone else adding a for(i...) loop in between.) Do whatever you think looks nice.

irc_commands.c
Wilm0r: Yup, I figured that after more careful reading. Also I then realised that that's how SASL works from having implemented the client version of this for the Jabber module .... 9 years ago now?

irc_commands.c
dequis: Oh fuck I already fixed this stuff the other day and forgot to push. PUSHAN

irc_commands.c
Wilm0r: Hrm, ok, my memory sucks apparently. I was convinced BitlBee was more paranoid but indeeed not.

ns identify -noload bla
:root!root@localhost PRIVMSG &bitlbee :The nick is (probably) not registered
nick wilmer          
:wilm00r!w@localhost NICK wilmer
ns identify -noload bla
:root!root@localhost PRIVMSG &bitlbee :Incorrect password

irc_commands.c
Wilm0r: Ah, gotcha. Yeah, maybe the client will still show the wrong msg then. Meh, who cares.

Wilm0r:
Re a comment further up: With the current storage subsystem (hashed passwords) I don't think we'll be able to support anything other than PLAIN. :-(

Supporting SSL cert auth eventually (if BitlBee gets native SSL) would I guess be doable with a small storage tweak if we ever think that'd be worth the trouble.

dequis:
>Supporting SSL cert auth eventually (if BitlBee gets native SSL) would I guess be doable with a small storage tweak if we ever think that'd be worth the trouble.

How? We need something to encrypt IM passwords.

Wilm0r:
> How? We need something to encrypt IM passwords.

HEh, good point. Forget what I said. :-)

grawity:
> How? We need something to encrypt IM passwords.

Couldn't the decryption passphrase be asked for separately if non-PLAIN auth was used?

dequis:
>Couldn't the decryption passphrase be asked for separately if non-PLAIN auth was used?

Yeah, quoting myself from above:

>unless the user's account is somehow configured to require certfp auth first, and only accept the identify password later through some other method. I guess someone could consider that an extra layer of security. I'm not sure if it's worth bothering.

(I asked "How?" hoping he had a better idea)

root_commands.c
dequis: Done.

PR #47, opened 2015-09-11
IRCv3 capability negotiation

dequis:
Specs: [capability-negotiation-3.1](http://ircv3.net/specs/core/capability-negotiation-3.1.html) and [capability-negotiation-3.2](http://ircv3.net/specs/core/capability-negotiation-3.2.html)

The 3.2 spec is important because it deprecates several rarely used 3.1 features. The 3.1 spec is a mess. This is compliant with 3.2 by not doing anything - it doesn't need multiline replies or capabilities with values. Implementing sasl-3.2 changes that (it requires values), but that one is optional.

Should have PR'd this before #46, but, whatever. It's old code, I think jgeboski already said it looks good. Don't think I've heard from wilmer about this.


irc_cap.c
```diff
@@ -0,0 +1,174 @@
...
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License with
+  the Debian GNU/Linux distribution in /usr/share/common-licenses/GPL;
+  if not, write to the Free Software Foundation, Inc., 51 Franklin St.,
+  Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
```
Wilm0r: Can you add a link to the "RFC" here and maybe in other places where it'd be helpful? (For example with anchor links if possible.) I'll leave the latter to your judgement, mostly care about at least a link here.

I'll continue reading and see if I can review this without having to search for the doc myself now. :-)

irc_cap.c
```diff
@@ -0,0 +1,174 @@
...
+	{"foo", CAP_FOO},
+	{"bar", CAP_BAR},
+	{NULL},
+};
+
+static irc_cap_flag_t cap_flag_from_string(char *cap_name)
+{
+	int i;
+
+	if (!cap_name && !cap_name[0]) {
```
Wilm0r: Seems wrong?

irc_cap.c
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+	return 0;
+}
+
+static gboolean irc_cmd_cap_req(irc_t *irc, char *caps)
+{
+	int i;
+	char *lower = NULL;
+        char **split = NULL;
```
Wilm0r: Spaces instead of tab?

irc_cap.c
```diff
@@ -0,0 +1,174 @@
...
+			continue;   /* skip empty items (consecutive spaces) */
+		}
+
+		remove = (split[i][0] == '-');
+		flag = cap_flag_from_string(split[i]);
+		
+		if (!flag || (remove && !(irc->caps & flag))) {
+			/* unsupported cap, or removing something that isn't there */
+			g_strfreev(split);
+			return FALSE;
```
Wilm0r: You abort the loop once you find one invalid flag?

irc_cap.c
```diff
@@ -0,0 +1,174 @@
...
+{
+	int i;
+	GString *str = g_string_sized_new(256);
+
+	for (i = 0; supported_caps[i].name; i++) {
+		if (i != 0) {
+			g_string_append_c(str, ' ');
+		}
+		g_string_append(str, supported_caps[i].name);
+	}
```
Wilm0r: glib doesn't have a strjoin BTW?

irc_cap.c
```diff
@@ -0,0 +1,174 @@
...
+	if (g_strcasecmp(cmd[1], "LS") == 0) {
+		irc_cmd_cap_ls(irc, cmd[2]);
+
+	} else if (g_strcasecmp(cmd[1], "LIST") == 0) {
+		irc_cmd_cap_list(irc);
+
+	} else if (g_strcasecmp(cmd[1], "REQ") == 0) {
+		gboolean ack = irc_cmd_cap_req(irc, cmd[2]);
+
+		irc_send_cap(irc, ack ? "ACK" : "NAK", cmd[2] ? : "");
```
Wilm0r: So the spec is that you do this during login, you (IRC client) use CAP REQ to list all the capabilities you request upport for, the server will hopefully ACK, and once you're happy you send CAP END and you log in (either that's it or you then send USER and NICK)?

irc_cap.c
dequis: Well, I don't really pay attention to the emails so you could get away with removing this comment and i might not see it. Maybe.

irc_cap.c
dequis: Whoa, I didn't know I was capable of making this kind of mistake (it should be pretty damn obvious and hard to make with my editor configs)

Wilm0r:
Looks fine. Just wondering whether it'd be easier to have for example a hash table tracking all the negotiated caps. It simplifies code a lot, lets you kill some pretty boring helper functions and for example listing all existing and/or supported caps would just be a simple strjoin. (Not that that's really interesting because it's not a very commonly done operation.)

How often would you need to do capability checks? A hash lookup is more expensive than a bitwise & of course, I just wonder whether we need to care.

irc_cap.c
dequis: It's in the PR description

Wilm0r: Sorry, ok, I didn't see that. But the PR/description is gone once this is submitted, the code is what people will keep looking at.

dequis:
>How often would you need to do capability checks? A hash lookup is more expensive than a bitwise & of course, I just wonder whether we need to care.

That's one question I was expecting! Very roughly, I'd expect to have 1-10 checks per sent message, 1-10 checks per received message. Maaaaybe not worth it. I just like the feeling of lightweightness of the bitwise &. Not the most rational of the decisions. But unless we get to 32 caps (or dangerously close to it), it seems harmless.

In contrast, some stuff like display_timestamps happens every message, and uses set_getbool, which.. is a linked list lookup, in the global settings. It also happens to be equivalent to the ircv3 server-time, so uh, yeah.

dequis:
>It simplifies code a lot, lets you kill some pretty boring helper functions and for example listing all existing and/or supported caps would just be a simple strjoin. (Not that that's really interesting because it's not a very commonly done operation.)

An earlier version of this had something like a strjoin, actually. Don't quite remember what data structure I was using back then. But I went with this mostly because cap-3.2 adds capabilities with values, for example `sasl=PLAIN` to list the mechanisms, and it's going to be easier to add values with a plain old for loop.

irc_cap.c
dequis: Sorta. Client sends CAP LS, USER, NICK (and maybe PASS) together. For servers that support this, registration is held until the client says CAP END. For servers that don't, they just reply with 001 and continue as usual.

irc_cap.c
dequis: Added in the comment block before the license text

```c
/* IRCv3 CAP command
 *
 * Specs:
 *  - v3.1: http://ircv3.net/specs/core/capability-negotiation-3.1.html
 *  - v3.2: http://ircv3.net/specs/core/capability-negotiation-3.2.html
 *
 * */
```

Wilm0r:
That's a lot of checks. Still not an interesting performance hit likely yes, but meh, ok, you've written this already so why not.

I wonder whether it would be nicer to just have a string->int hash but really it all works for me. Sheep it.

dequis:
>Sheep it.

Yay! :sheep:

Squashed last few tiny fixes into the last commit, since they are only relevant that one. Merging this one, next is #46 (sasl) but will wait for your confirmation before merging it.

PR #48, opened 2015-09-12
IRCv3 multi-prefix... but mostly just adding prefixes to WHO

dequis:
We can't actually have multiple prefixes internally, so the only thing
missing for multi-prefix compliance is actually having the prefix in the
WHO reply, which is a rfc1459 thing.

Note to future self: check irc logs for the implementation I threw away.
The one that actually handled multiple prefixes. I hope that's useful.

------

Building a house of cards of pull requests.

This one (`wip/ircv3+sasl+multi-prefix`) applies against `wip/irc3+sasl`.

I suspect branches aren't meant to be used like this.

----

This one would normally not require any review, but it's 5am and the `char status_prefix[3]` thing is probably a bit too clever. It works, and I believe it's sane, but valgrind isn't going to tell me if i'm doing dumb stuff because it's in the stack.

### DIFF
--- a/irc_send.c
+++ b/irc_send.c
@@ -213,13 +213,8 @@ void irc_send_names(irc_channel_t *ic)
 			*namelist = 0;
 		}
 
-		if (icu->flags & IRC_CHANNEL_USER_OP) {
-			strcat(namelist, "@");
-		} else if (icu->flags & IRC_CHANNEL_USER_HALFOP) {
-			strcat(namelist, "%");
-		} else if (icu->flags & IRC_CHANNEL_USER_VOICE) {
-			strcat(namelist, "+");
-		}
+		namelist[strlen(namelist) + 1] = '\0';
+		namelist[strlen(namelist)] = irc_channel_user_get_prefix(icu);
 
 		strcat(namelist, iu->nick);
 		strcat(namelist, " ");
@@ -293,15 +288,26 @@ void irc_send_who(irc_t *irc, GSList *l, const char *channel)
 	gboolean is_channel = strchr(CTYPES, channel[0]) != NULL;
 
 	while (l) {
-		irc_user_t *iu = l->data;
+		irc_user_t *iu;
+
+		/* Null terminated string with three chars, respectively:
+		 * { <H|G>, <@|%|+|\0>, \0 } */
+		char status_prefix[3] = {0};
+
+		/* rfc1459 doesn't mention this: G means gone, H means here */
+		status_prefix[0] = iu->flags & IRC_USER_AWAY ? 'G' : 'H';
+
 		if (is_channel) {
-			iu = ((irc_channel_user_t *) iu)->iu;
+			irc_channel_user_t *icu = l->data;
+			status_prefix[1] = irc_channel_user_get_prefix(icu);
+			iu = icu->iu;
+		} else {
+			iu = l->data;
 		}
-		/* TODO(wilmer): Restore away/channel information here */
-		irc_send_num(irc, 352, "%s %s %s %s %s %c :0 %s",
+
+		irc_send_num(irc, 352, "%s %s %s %s %s %s :0 %s",
 		             is_channel ? channel : "*", iu->user, iu->host, irc->root->host,
-		             iu->nick, iu->flags & IRC_USER_AWAY ? 'G' : 'H',
-		             iu->fullname);
+		             iu->nick, status_prefix, iu->fullname);
 		l = l->next;
 	}
 
--- a/irc.h
+++ b/irc.h
@@ -68,6 +68,7 @@ typedef enum {
 
 typedef enum {
 	CAP_SASL = (1 << 0),
+	CAP_MULTI_PREFIX = (1 << 1),
 } irc_cap_flag_t;
 
 struct irc_user;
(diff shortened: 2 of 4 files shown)


irc_send.c
```diff
@@ -293,15 +287,26 @@ void irc_send_who(irc_t *irc, GSList *l, const char *channel)
 	gboolean is_channel = strchr(CTYPES, channel[0]) != NULL;
 
 	while (l) {
-		irc_user_t *iu = l->data;
+		irc_user_t *iu;
+
+		/* Null terminated string with three chars, respectively:
+		 * { <H|G>, <@|%|+|\0>, \0 } */
+		char status_prefix[3] = {0};
```
Xe: `multi-prefix` shows all the prefixes in here:

```
JOIN #foo
:a!~a@127.0.y.p JOIN #foo
:cadance.int MODE #foo +nt
:cadance.int 353 a = #foo :@a
:cadance.int 366 a #foo :End of /NAMES list.
MODE #foo +h a
:a!~a@127.0.y.p MODE #foo +h a
WHO #foo
:cadance.int 352 a #foo ~a 127.0.y.p cadance.int a H@% :0 a
:cadance.int 315 a #foo :End of /WHO list.
```

dequis: Yeah, well, quoting commit log:

>We can't actually have multiple prefixes internally

Modes/prefixes in bitlbee are actually just a 1:1 mapping from user status (online, away, special, offline) to a single character, determined by the `show_users` channel setting (which is a string like `online@,special%,away+,offline`, defaults to `online+,away`). That's in the case of control channels.

Chat channels don't actually support modes other than showing yourself and the root user as @. Even if they supported modes (to show who's a moderator in a groupchat, and i really want to implement this soonish), I don't think any IM protocols other than IRC support multiple statuses.

An earlier version of this patch had actual support for multiple prefixes:

http://dump.dequis.org/XENN4.txt (oh look, it almost has your name as the hash, neat)

...but i threw it away once i got to test it, because I couldn't find any way to set both flags at the same time. 

Either way, thanks for looking!

Xe: XMPP does, as does skype (but skype doesn't expose this)

dequis: hmmm, from xep-0045:

>MUC also defines two different associations: long-lived affiliations and session-specific roles. These two association types are distinct from each other in MUC, since an affiliation lasts across visits, while a role lasts only for the duration of a visit. In addition, there is no one-to-one correspondence between roles and affiliations; for example, someone who is not affiliated with a room may be a (temporary) moderator, and a member may be a participant or a visitor in a moderated room.

Roles: Moderator, participant, visitor, none

Affiliations: Owner, admin, member, outcast, none

And affiliations define the default role when joining a channel (owner and admin become moderators, members are always participant, none depends on how open the channel is, outcast is always banned)

I'd say roles are like modes, and affiliations are like chanserv access levels. Moderator maps to @, the rest map to no mode.

Anyway the groupchat code doesn't even support this at all yet. And when mapping this part of IM services, i'd rather go for the simplest solution

irc_send.c
```diff
@@ -213,13 +213,7 @@ void irc_send_names(irc_channel_t *ic)
...
 		}
 
-		if (icu->flags & IRC_CHANNEL_USER_OP) {
-			strcat(namelist, "@");
-		} else if (icu->flags & IRC_CHANNEL_USER_HALFOP) {
-			strcat(namelist, "%");
-		} else if (icu->flags & IRC_CHANNEL_USER_VOICE) {
-			strcat(namelist, "+");
-		}
+		namelist[strlen(namelist)] = irc_channel_user_get_prefix(icu);
```
Wilm0r: This does terrify me somewhat though the string should I think be fully \0-initialised by assigning "" to it, does glib maybe have an append_char function?

dequis: Hmm, I know `array = {0}` is be guaranteed to initialize a whole array to 0 [citation needed], and technically "" is the same thing but... yeah, not sure.

Same thing applies to the old strcat() code though.

Could certainly use GStrings in this code to make it cleaner, if we don't mind allocating some more stuff in the heap instead of the stack. Probably not worth changing though.

irc_send.c
```diff
@@ -293,15 +287,26 @@ void irc_send_who(irc_t *irc, GSList *l, const char *channel)
...
+			status_prefix[1] = irc_channel_user_get_prefix(icu);
+			iu = icu->iu;
+		} else {
+			iu = l->data;
 		}
-		/* TODO(wilmer): Restore away/channel information here */
-		irc_send_num(irc, 352, "%s %s %s %s %s %c :0 %s",
+
+		/* rfc1459 doesn't mention this: G means gone, H means here */
+		status_prefix[0] = iu->flags & IRC_USER_AWAY ? 'G' : 'H';
```
Wilm0r: I guess this is fine yeah. Could consider moving this one up so that the string is not filled in reverse order.

irc_send.c
Wilm0r: Actually https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html is not about GStrings. But it doesn't even have a char-append function so never mind.

strcat() should be safe since I do assume it would add a new trailing \0 which is what I was missing here. But as said I think this should not be necessary. (Though I don't remember whether the function does other naughty things that don't do that.)

irc_send.c
dequis: That's actually reformatted from your old code, `iu->flags & IRC_USER_AWAY ? 'G' : 'H'`, a couple of lines below. I just added the comment (because they confused me) and removed the TODO (which wasn't valid anymore)

re: moving up, good idea.

irc_send.c
dequis: OOOHHHhhhh i get your point now. Yeah, my code is certainly doing it wrong here. If the names list fills the 385 byte buffer, it's sent through and only the first byte is set to 0. Starting from the second pass, the buffer definitely holds garbage.

How about

```c
namelist[strlen(namelist) + 1] = '\0'
namelist[strlen(namelist)] = irc_channel_user_get_prefix(icu);
```

dequis:
Cherry picked this thing into 687ec880201e4872a9abf72f5efe135164fc349b 

>02:24 < dx> wilmer: going to go ahead and merge multi-prefix into master, since it's highly unlikely that you'll find more issues with it

Also because this chained PR thing i made is dumb

PR #49, opened 2015-09-13
IRC self-message support (messages sent by yourself from other clients)

dequis:
This adds an OPT_SELFMESSAGE flag that can be passed to imcb_buddy_msg()
or imcb_chat_msg() to indicate that the protocol knows that the message
being sent is a self message.

This needs to be explicit since the old behavior is to silently drop
these messages, which also removed server echoes.

This commit doesn't break API/ABI, the flags parameters that were added
are all internal (between protocols and UI code)

On the irc protocol side, the situation isn't very nice, since some
clients put these messages in the wrong window. Irssi, hexchat and mirc
get this wrong. Irssi 0.8.18 has a fix for it, and the others have
scripts to patch it.

But meanwhile, there's a "self_messages" global setting that lets users
disable this, or get them as normal messages / notices with a "->"
prefix, which loosely imitates the workaround used by the ZNC
"privmsg_prefix" module.

### DIFF
--- a/doc/user-guide/commands.xml
+++ b/doc/user-guide/commands.xml
@@ -1403,6 +1403,33 @@
 		</description>
 	</bitlbee-setting>
 
+	<bitlbee-setting name="self_messages" type="string" scope="global">
+		<default>true</default>
+		<possible-values>true, false, prefix, prefix_notice</possible-values>
+
+		<description>
+			<para>
+				Change this setting to customize how (or whether) to show self-messages, which are messages sent by yourself from other locations (for example, mobile clients), for IM protocols that support it.
+			</para>
+
+			<para>
+				When this is set to "true", it will send those messages in the "standard" way, which is a PRIVMSG with source and target fields swapped.
+			</para>
+			
+			<para>
+				Since this isn't very well supported by some clients (the messages might appear in the wrong window), you can set it to "prefix" to show them as a normal message prefixed with "-> ", or use "prefix_notice" which is the same thing but with a NOTICE instead.
+			</para>
+
+			<para>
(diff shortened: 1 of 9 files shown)


Xe:
It's worth noting that weechat has the correct behavior out of the box.

dequis:
Yeah I only mentioned the broken ones. I think I'll make a bitlbee wiki page equivalent to http://wiki.znc.in/Privmsg

dequis:
bitlbee-facebook implementation: https://github.com/jgeboski/bitlbee-facebook/pull/44 (waiting on this PR)

MSN private messages and jabber groupchat implementation: https://github.com/bitlbee/bitlbee/tree/self-messages+protocols (branch on top of this PR)

(MSN has no groupchats atm, and jabber private messages need the carbons patch, soon™)

chauffer:
:+1:

Wilm0r:
Fun. I would only add more clearly to the doc what "true" does.

Also, you're saying that IRC channels just work. You've tried this in many clients? I'd at least rephrase that to "has not caused issues in any clients so far" or something because I'd be surprised if not a single client fucked that up. (But we can deal with it code-wise once that corner case comes up.. if at all.)

dequis:
>Fun. I would only add more clearly to the doc what "true" does.

Sure.

>You've tried this in many clients?

To be honest, no. Mostly going by what the ZNC people documented about this topic in http://wiki.znc.in/Privmsg

>I'd be surprised if not a single client fucked that up

Channel messages just work because all clients have to read both source and target, while with private messages they could get away with only reading the source and assume the target is always the same.

But yeah, some client may have a misguided attempt to sanitize inputs or something like that.

PR #51, opened 2015-10-07
Add missing space in configure (otr)

Joschasa:
The missing space resulted in EFLAGS+=-lotr-lgcrypt, which lead to a linker error.

### DIFF
--- a/configure
+++ b/configure
@@ -677,15 +677,15 @@ fi
 if [ "$otr" = 1 ]; then
 	# BI == built-in
 	echo '#define OTR_BI' >> config.h
-	echo "EFLAGS+=$($PKG_CONFIG --libs libotr)$(libgcrypt-config --libs)" >> Makefile.settings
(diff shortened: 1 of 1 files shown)


dequis:
Weird, it works fine for me (and works fine for travis too)

```
$ echo $(pkg-config --libs libotr)$(libgcrypt-config --libs)
-lotr -lgcrypt -lgpg-error
```

What platform?

Joschasa:
Arch linux

dequis:
Same here. Wat.

Joschasa:
Tested on two machines, the space is missing on both.

The section for gnutls uses a space between the commands:
https://github.com/Joschasa/bitlbee/blob/fixotrconfigure/configure#L374

dequis:
I think the patch it's okay, no need to justify it, really, i just don't get why it works for me.

dequis:
Oh. Upgrading from pkg-config 0.28 to 0.29 did that. From the changelog:

>Strip trailing space from --cflags/--libs output

Welp.

dequis:
Merged as 872e0176b87ad859a66a1ccbe4f07211ccddb11f

Thanks!

PR #53, opened 2015-10-07
Allow NULL callback functions in http_dorequest

sm00th:
Check callback function supplied to http_dorequest and only run it if it is
not NULL.

While it is not the usual case there are some times when there is no need to
check the results of a http request. Using a NULL pointer is much more
convenient than creating noop functions.

### DIFF
--- a/lib/http_client.c
+++ b/lib/http_client.c
@@ -161,7 +161,9 @@ error:
 		req->status_string = g_strdup("Error while writing HTTP request");
 	}
 
-	req->func(req);
+	if (req->func != NULL) {
+		req->func(req);
+	}
 	http_free(req);
 	return FALSE;
 }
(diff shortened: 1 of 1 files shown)


jgeboski:
This would be useful for [bitlbee-facebook](https://github.com/jgeboski/bitlbee-facebook/blob/635c9cef534f9b469e1adfca0b5f091789eaef1c/facebook/facebook-http.c#L99) and [bitlbee-steam](https://github.com/jgeboski/bitlbee-steam/blob/011375b2d3c67c15d51ca203de0ecaab3b4b7587/steam/steam-http.c#L360).

dequis:
Merged as 0b0bb4c68a1c87b8f77f266ea8e836758203f53e

Thanks!

PR #55, opened 2015-11-04
Fix https://bugs.bitlbee.org/bitlbee/ticket/415

wingo:
When I would join a non-anonymous conference room cafe@foo.comm the IRC
nicks were being derived from the participants' jabber ID, and not the
handle with which they joined the room. For example one user in the
roster would be "agarcia" and not "berto". The "berto" name came in the
roster for the room, i.e. a jabber ID of cafe@foo.com/berto.  The name
"berto" doesn't appear anywhere else.  The jabber server is Openfire
3.6.4.

The reason this information wasn't getting propagated correctly appears
to be twofold.  One, the nick_hint wasn't getting set on the buddies for
non-anonymous chats.  Two, setting the nick hint would bail out if the
user was already logged in -- but of course the user is logged in,
that's why I can see them.

This patch fixes both issues.

### DIFF
--- a/irc_im.c
+++ b/irc_im.c
@@ -402,11 +402,6 @@ static gboolean bee_irc_user_nick_update(irc_user_t *iu)
 	bee_user_t *bu = iu->bu;
 	char *newnick;
 
-	if (bu->flags & BEE_USER_ONLINE) {
-		/* Ignore if the user is visible already. */
-		return TRUE;
-	}
-
 	if (nick_saved(bu)) {
 		/* The user already assigned a nickname to this person. */
 		return TRUE;
--- a/protocols/jabber/conference.c
+++ b/protocols/jabber/conference.c
@@ -312,7 +312,6 @@ void jabber_chat_pkt_presence(struct im_connection *ic, struct jabber_buddy *bud
 			/* If JIDs are anonymized, add them to the local
 			   list for the duration of this chat. */
 			imcb_add_buddy(ic, bud->ext_jid, NULL);
-			imcb_buddy_nick_hint(ic, bud->ext_jid, bud->resource);
 		}
 
 		if (bud == jc->me && jc->invite != NULL) {
@@ -327,6 +326,7 @@ void jabber_chat_pkt_presence(struct im_connection *ic, struct jabber_buddy *bud
 		if (s) {
 			*s = 0; /* Should NEVER be NULL, but who knows... */
 		}
+		imcb_buddy_nick_hint(ic, bud->ext_jid, bud->resource);
 		imcb_chat_add_buddy(chat, bud->ext_jid);
 		if (s) {
 			*s = '/';


dequis:
Thanks for looking into this issue!

> Two, setting the nick hint would bail out if the user was already logged in -- but of course the user is logged in, that's why I can see them.

This change doesn't sound right. It also doesn't explain why it works for anonymous MUCs.

I'll take a deeper look at this patch later, obviously just skimming it now isn't enough.

andywingo:
One difference for anonymous MUCS is that at the point the nick is set in that case, the ext_id contains the resource.  Later on in the chat_pkt_presence the slash is nulled out.  In fact in the non-anonymous case we have to set the nick without the user resource (not the same as the resource in the chat JID) otherwise it doesn't work.

dequis:
Just dumping notes here:

d323394 is the original commit that added the online check when renaming

d06eabf moved it to imcb_buddy_nick_hint() with the following comment:

    /* Only do this if the person isn't online yet (which should
       be the case if we just added it) and if the user hasn't
       assigned a nickname to this buddy already. */

6ef9065 is (almost) its current form, moved from nogaim to irc_im, with a slightly different online check and a shorter comment

    /* Ignore if the user is visible already. */`

The current `bee_irc_user_nick_update()` (where the online check is being removed) calls `irc_user_set_nick()`, which loops over the joined channels and calls `irc_send_nick()` if the following condition is met:

    /* Send a NICK update if we're renaming our user, or someone
       who's in the same channel like our user. */

Still can't see a reason for that check to exist, although it looks like an important part of its contract. As in, calling nick_hint in the wrong moment is guaranteed to be a no-op - maybe another protocol plugin depends on this behavior? On the other hand it sounds desirable to change nicks sometimes (I've actually run into this check before when trying to handle twitter nick changes)

One possible problem I can imagine is joining several non-anonymous MUCs where a person has different nicks in each, but that might be an unavoidable side effect of supporting this feature at all.

dequis:
lol `bee_irc_user_nick_reset()`

```c
	/* In this case, pretend the user is offline. */
	if ((online = bu->flags & BEE_USER_ONLINE)) {
		bu->flags &= ~BEE_USER_ONLINE;
	}

	nick_del(bu);
	bee_irc_user_nick_update(iu);

	bu->flags |= online;
```

@Wilm0r any idea why `bee_irc_user_nick_update()` ignores nick hints if the user already has BEE_USER_ONLINE? ([irc_im.c:405][1])

[1]: https://github.com/bitlbee/bitlbee/blob/345577b/irc_im.c#L405

andywingo:
Thanks for looking into this issue, @dequis :)

You say:

"One possible problem I can imagine is joining several non-anonymous MUCs where a person has different nicks in each, but that might be an unavoidable side effect of supporting this feature at all."

And indeed this is the case.  As a user you can get around this issue via explicitly giving them a nick, AFAIU.  But given the choice between them appearing with a nick unrelated to the one they have in the chat and possible nick churn, I definitely prefer to know that if I write "berto: ping" in a channel that it actually let berto know I want to talk to him, and not "agarcia" just because agarcia was part of his jabber id :)

dequis:
This also fixes the translate_to_nicks setting, so the following just works:

* `set nick_format %-@nick|%account`
* `user1` joins #muc with default nick
* `user2` joins #muc with nick set to `a s d`
* `user1` sees `asd|user1` join
* Also as a side effect of this PR, `user2` changes nick in the control channel of `user1`
* `user1` writes `asd|user1: hi` in #muc
* `user2` receives `a s d: hi`

And also after renaming:

* `user1` does `rename asd|user1 user2|user1` in the control channel
* `user1` says `user2|user1: hi` in #muc
* `user2` receives `a s d: hi`

>I definitely prefer to know that if I write "berto: ping" in a channel that it actually let berto know I want to talk to him, and not "agarcia" just because agarcia was part of his jabber id :)

Now, this looks like a place where translate_to_nicks could do something interesting.

Buuut nope. It does what you said.

* `user1` joins #muc2 with default nick
* `user2` joins #muc2 with nick set to `s d f`
* `user1` sees `user2|user1` join (would have been `sdf|user1`, with a global nick change again, but the rename takes priority)
* `user1` says `user2|user1: hi` in #muc2
* `user2` receives `s d f: hi`
* `user1` says `user2|user1: hi` in #muc (the old one)
* `user2` receives `s d f: hi` (instead of `a s d: hi`)

Less than ideal but still much better, and as you say, probably even acceptable.

dequis:
While i'm taking my time pondering whether i like this change or not, could you change the commit message to have something more meaningful in the subject line? Like "Fix nicks in non-anonymous jabber groupchats", and move the trac link to the body of the message. Just `git commit --amend` and `git push --force` to the same branch. Thanks!

Wilm0r:
> One possible problem I can imagine is joining several non-anonymous MUCs where a person has different nicks in each, but that might be an unavoidable side effect of supporting this feature at all.

Yep that's a problem and IMHO why, if one wants to fix this, they should just make BitlBee use the JID handles in channels. Maybe expose the real JIDs somehow if you can, but not the external one since it's just going to be trouble like above. It'd also undo any manual renames you may have done for the contact in your own list.

The reason nick hints are ignored once the contact is online is to avoid (constant) nick changes. Once online the nick is set in stone unless the user changes it themselves. This works fine since modules are passing the nick hint info (and contact list details in general) before presence.

dequis:
>just make BitlBee use the JID handles in channels

You mean the "nick=username@domain" fake ext JIDs? The problem with that is that underscores. Lots of underscores. Which is the same thing that would happen if you had several anonymous channels with the same person, but that's currently avoided if all those channels are non-anonymous. It would make things really annoying in servers that are structured like hipchat - several chat channels, everyone in everyone elses's contact list, no anonymity at all.

>It'd also undo any manual renames you may have done for the contact in your own list.

It doesn't! Manual renames take priority. That part seems to be working fine.

>Once online the nick is set in stone unless the user changes it themselves. This works fine since modules are passing the nick hint info (and contact list details in general) before presence.

I'd really like to have a way to force nick changes from the protocol, though. Maaaaybe nick hints are not the right way to do it, and a new imcb function may be needed.

But, for example, there's no way to handle a twitter user changing their username.

Or in jabber itself, where changing nicks in a MUC is a leave/rejoin, it would be nice if an actual nick change could be sent to clients. And there's no api for this, other than what this PR changes.

andywingo:
Fixed the commit message, no code changes.

Wilm0r:
> You mean the "nick=username@domain" fake ext JIDs? The problem with that is that underscores. Lots of underscores.

That's true, but I don't know whether a full merge is the right solution. As you say a manually-set nick will be used over in-channel nicks but that will still leave people unhappy about this. Plus no guarantees that person X will have the same nick in all channels/rooms they are in.

An idea to partially fix this is to have a room- and a plain-contact- version of people but only one room-contact even if the person is in multiple rooms. But meh, I'd go for the underscores, you'll still have that problem anyway in anonymous rooms (ah, as you also pointed out already). And really there is no nice way to solve this problem with Jabber having per-room namespaces and IRC per-server. (Well unless you want to do the really horrible thing of having additional fake IRC servers for each room!)

I do wonder what's the average number of rooms people are in. I would be surprised if many people were in more than 1 or maybe 2 rooms?

> Or in jabber itself, where changing nicks in a MUC is a leave/rejoin

IIRC BTW this is what it really looks like on the wire as well and you actually have to read some extra fields to figure out it's actually a rename?

And yeah, handling renames on Twitter would be nice although that's rather obscure. But yeah, something stronger than a nick_hint would be nice. BTW in case of Twitter it's not just the nick that changes but also the handle, so that's a little different anyway.

dequis:
Other ideas:

* A setting to make this behavior optional (allow renames that affect a contact globally, but only if the user agrees to have it that way)
* A setting to make non-anonymous channels behave like anonymous ones, with fake ext JIDs for everyone and underscores.
* As a complement to any other solution:
    * Would be nice if translate_to_nicks looked up the correct room nick, so regardless of what the user sees, the correct thing is sent over the wire
    * Also would be nice to send additional fields over /whois (sorta like info), including information about each roomnick.
    * Maybe this is a good opportunity to implement the very old idea of a rename-like command to merge contacts (across different accounts and channels). Except that this change also makes it desirable to have several nicks for more-or-less the same person. That can get confusing fast.

andywingo:
On Sat 07 Nov 2015 03:10, Wilmer van der Gaast <notifications@github.com> writes:

> I do wonder what's the average number of rooms people are in. I would be
> surprised if many people were in more than 1 or maybe 2 rooms?

FWIW I sit in around 8 rooms, though there are more I could join in if I
like.  Everyone else in my company (around 42 people) does the same.

I really want their IRC nick to be the nick with which they join the
rooms.

Andy

Wilm0r:
Wow, 8 rooms. Thanks for that very helpful datapoint. Underscores would
get very very awful then.

I think the best approach (and I must add that I am typing this via
e-mail without inet access so I have no idea what is in the actual
chnage) would maybe be a setting that lets you choose between three choices:

A) Current behaviour.
B) Your behaviour which is current behaviour except nicknames in rooms
are used. If a person has different nicks in different rooms that will
not work.
C) Make non-anon rooms behave like anon ones -> underscores.

And maybe there could be another:

D) roomname@servername/nick becomes something like
nick@"ROOM".servername (so without the roomname, just literal "ROOM")
which will mean nicks will (mostly) not have (multiple) underscores,
they will match what people had in the room even when people can't make
their minds up, and they will be separated from the contact list nicks.

andywingo:
Ping :)  I am pretty sure that if people are in multiple non-anonymous jabber chat rooms, they want the same nick each place, like in IRC.  (Of course, maybe not, but given a choice between (B) and (A) for them, (B) is pretty clearly the right choice.)

(A), (C), and (D) do not solve my use case.  When I type "berto: hey" in a room I want berto to be notified (on his end), and that only happens if I use the nick that _he_ has configured.  Also I need that room to show up as needing attention, in his IM client.  That's what (B) does.

I humbly submit that my patch is pretty close to the right thing, and at any event is closer than where we are now, and that it should just be applied as-is :)

Wilm0r:
So I would definitely prefer any improvement to be a setting instead of just an enforced change in behaviour. However an advantage of your change is that at least if the user has already used rename to assign nicks to contacts, those nicks will be preserved over whatever a room may have set.

And implementing all four options as mentioned above would, I will admit, likely be over the top.

andywingo:
I'm OK with the option but I wonder where to put it and what to name it, given there are two changes :)  I guess from the user's POV the option is whether non-anonymous jabber chats should override the buddy's global nick and therefore it should be in the jabber code.  That would imply that the other change to make nick_hint work when the buddy is logged-in should be enabled regardless.  Sound about right to you @Wilm0r ?

ricardoamaro:
+1 for this fix

andywingo:
Ping to @Wilm0r for feedback/confirmation regarding where to put the option, if an option is needed, as per my last message.  Thanks :)

dequis:
Okay I was meaning to write this a while ago, sorry for the delay, but here: df701918487fcebd66f46dba0e2d147d72a42eb7

It's a variant of `imcb_buddy_nick_hint()` that doesn't care if the user is already online. Usage is the same:

```c
imcb_buddy_nick_change(ic, bud->ext_jid, bud->resource);
```

This takes care of the irc_im.c part of your patch.

For the rest we should figure out a decent setting name to change nicks globally in non-anonymous channels. Naming is hard.

andywingo:
Happy new year :)

Thanks @dequis for your patch.  I am a little confused about where this is going though as you don't seem to have merged your patch into master yet.  What is the next step here?  I'm happy to rewrite the nick-changing setting once I can test that it works with whatever nick-update mechanism that lands.

wingo:
Ping :)  Still wondering about the next step.

dequis:
Hey sorry for the silence, I was thinking about merging this after the 3.4.2 release (the master branch is in an informal bugfix only freeze) but it got unexpectedly delayed. I already had this merged locally together with a few other patches that complete this, I think i'm just going to dust off the develop branch and throw everything in there. Unless 3.4.2 happens.

wingo:
Thanks for the update, good luck with the release :)

dequis:
Pushed my local commits to the `dx/kitchen-sink` branch, 7da82f7ed729b8c9308ced6a6d95bd3f81bf943b is the one that implements this stuff.

I'm open to name ideas and changes to the help text. I picked the first name that came to my mind to avoid getting stuck with that and actually write the thing.

wingo:
Thank you @dequis!  I will give it a try soon.

dequis:
3.4.2 is out, master is unfrozen, rebased and merged kitchen sink, and this became a42fda42abad6af64ac9a905856ee9a3095954cd and 3320d6d9868729ce1d7b5a90866554bc898a1fa3

(still open to name changes to the setting)

wingo:
I can confirm that it seems to work; thanks!

The one limitation is that since it's a channel-scoped setting I have to set it for all my channels (about 8 or 9 right now), and that if I forget to set it, then I have the problem I originally described.  In my use case, it's a setting that I always want to be on for all of my channels.  Does it not make sense to have a global setting?  I would be happy to make a patch, if that made sense.

PR #56, opened 2015-11-07
Do not use NOTICE for OTR messages for a given user

revmischa:
I cannot count how many times I've lost messages from someone because they ended their OTR session and the information was sent in a NOTICE in a user window. I literally cannot count, because I don't know how often it happens. This is really broken behavior and after years of it driving me mad I fixed it myself.

### DIFF
--- a/otr.c
+++ b/otr.c
@@ -1399,8 +1399,8 @@ void display_otr_message(void *opdata, ConnContext *ctx, const char *fmt, ...)
 	va_end(va);
 
 	if (u) {
-		/* display as a notice from this particular user */
-		irc_usernotice(u, "%s", msg);
+		/* just show this as a regular message */
+		irc_usermsg(u, "<<\002OTR\002>> %s", msg);
 	} else {
 		irc_rootmsg(irc, "[otr] %s", msg);
 	}


dequis:
...well that's one way of doing it.

I looked into changing this behavior before but got lost trying to re-inject the message through the usual methods. But of course it already deals with the irc layer directly, with `irc_usernotice`. I guess it's harmless to do it this this way.

revmischa:
P.S. it's tested
screen shot 2015-11-07 at 3 52 06 pm

dequis:
Why `irc_send_msg_f()` instead of  `irc_usermsg()`? The latter has the same signature as `irc_usernotice()`

revmischa:
To prepend the notice that it's an OTR system message

revmischa:
I have no idea if `irc_usernotice()` accepts formatted strings, I don't know anything about these libraries. I just wanted this fixed.

dequis:
Well, you have a `%s` right there in `irc_usernotice(u, "%s", msg);`. Change that and i'll merge it. 

(Bonus points for `git commit --amend` so i don't have to squash)

revmischa:
I hate git

dequis:
BONUS POINTS LOST

Cherry-picked the middle commit into 0a5d03c8aa25223d054012022caeecfebf7b9638

Added 3d365b497cc788ae330e9783799cee4ad2444368 as a """security""" measure (although most important system messages still use notices, so whatever)

dequis:
Also, thanks!

revmischa:
Thank you!

revmischa:
Message me let's try it out

PR #57, opened 2015-12-11
Remove nonexistent forward declaration

esainane:
imcb_find_buddy is mentioned only in nogaim.h, and is never defined anywhere.

This is misleading for plugin authors, who were probably looking for
(the actually implemented) imcb_buddy_by_handle instead.


dequis:
Thanks!

PR #58, opened 2016-02-01
Check global account away setting in WHO replies

digitalcircuit:
When sending a WHO reply, check if irc_user is the account user.  If so, set away status according to global account away configuration.  Otherwise, reply according to irc_user flags.  This fixes global away status not being shown in IRC clients when WHO polling is enabled, e.g. Quassel IRC.

Fixes:  https://bugs.bitlbee.org/ticket/1247

### DIFF
--- a/irc_send.c
+++ b/irc_send.c
@@ -347,8 +347,13 @@ void irc_send_who(irc_t *irc, GSList *l, const char *channel)
 			iu = l->data;
 		}
 
-		/* rfc1459 doesn't mention this: G means gone, H means here */
-		status_prefix[0] = iu->flags & IRC_USER_AWAY ? 'G' : 'H';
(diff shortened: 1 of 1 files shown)


dequis:
While I'm not a huge fan of the `set_getstr()` in here - i'd rather have it use flags for `irc->user` too - I'm going to merge it now since making it use flags requires a couple of non-trivial changes that I don't want to do right before 3.4.2 (the set callback doesn't have direct access to irc->user and bee->user is a lie)

digitalcircuit:
As discussed on IRC, I agree, the settings check isn't ideal.  I intended to avoid making non-trivial changes without first gaining a better understanding of Bitlbee's design.  However, time permitting, I'd be happy to try to help with designing or implementing other solutions post-3.4.2.  Thank you!

PR #59, opened 2016-02-04
Send IRCv3 away-notify after nicks join/part from IRC channel

digitalcircuit:
Move ```irc_send_away_notify``` below ```bee_irc_channel_update``` to delay sending the updated away-notify status until after any nicks have joined/quit.  Otherwise, some IRC clients will drop the status messages as they go to nicks that the client doesn't know about.


dequis:
Sorry for the delay with this one, but it's merged. Thanks! I'll rebase the 3.4.2 branch to include this stuff later.

PR #60, opened 2016-02-09
Change twitter url length to 23 instead of 22

sjefen6:
Warning, I have not tested this!

Looks like twitter changed the length of t.co urls again. This might fix the issue (looks like I am missing a step in compiling/installing). This is not the propper way to fix this issue. The correct way to fix it would be reading the length from https://dev.twitter.com/rest/reference/get/help/configuration


dequis:
Merged as 94562555d64001fc5517b328d1c8b8493dfd09cd (removed the "not tested" from the commit message)

Thanks!

PR #61, opened 2016-02-22
Support for locked-down configurations

seveas:
I operate a bitblee server that is locked down a bit more than normal. Not only
is registration disabled, the configs are also pregenerated (they actually come
from a different storage backend using ldap and mysql).

In this setup it is incredibly useful to be able to say "this part of the
settings may not change". Specifically, anything around authentication and the
addition/deletion of accounts needs to be locked down.

These two commits accomplish this in a configurable way:

- In the pregenerated configs, you can set an account to be locked by adding a
  locked="true" attribute. This disables changing the username/password/server
  and deleting the account
- A new configuration option is added to disable account creation.

### DIFF
--- a/conf.c
+++ b/conf.c
@@ -56,6 +56,7 @@ conf_t *conf_load(int argc, char *argv[])
 	conf->authmode = AUTHMODE_OPEN;
 	conf->auth_pass = NULL;
 	conf->oper_pass = NULL;
+	conf->allow_account_add = 1;
 	conf->configdir = g_strdup(CONFIG);
 	conf->plugindir = g_strdup(PLUGINDIR);
 	conf->pidfile = g_strdup(PIDFILE);
@@ -245,6 +246,12 @@ static int conf_loadini(conf_t *conf, char *file)
 			} else if (g_strcasecmp(ini->key, "operpassword") == 0) {
 				g_free(conf->oper_pass);
 				conf->oper_pass = g_strdup(ini->value);
+			} else if (g_strcasecmp(ini->key, "allowaccountadd") == 0) {
+				if (!is_bool(ini->value)) {
+					fprintf(stderr, "Invalid %s value: %s\n", ini->key, ini->value);
+					return 0;
+				}
+				conf->allow_account_add = bool2int(ini->value);
 			} else if (g_strcasecmp(ini->key, "hostname") == 0) {
 				g_free(conf->hostname);
 				conf->hostname = g_strdup(ini->value);
--- a/storage_xml.c
+++ b/storage_xml.c
@@ -85,7 +85,7 @@ static void handle_settings(struct xt_node *node, set_t **head)
 static xt_status handle_account(struct xt_node *node, gpointer data)
 {
 	struct xml_parsedata *xd = data;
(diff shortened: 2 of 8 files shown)


dequis:
Fancy. Skimmed, looks mostly ok.

Not sure if I like the idea of marking individual settings as locked down in the code of those protocols. It would make more sense to do that from the config file, and let admins specify how far they want to go with the locking. I guess you could do that lookup in set_add()

seveas:
I thought of allowing individual settings to be locked down, and could easily add that. But you'd still need the account lockdown to disallow deleting it, and it just made sense (to me at least) to then also lock down username/password/server.

dequis:
Yeah that part is fine, the problem is that this is limited to the protocols that are shipped as part of bitlbee. If it was done in the conf file, it would just work and would require less changes.

seveas:
Ok, ca91cf5 adds that for individual settings. I think I can rip out the server/port settings from 535fd59, now, but the username and password are not individual settings, so the support for 'account locked down means these settings can't change' cannot be removed completely.

dequis:
That's.. slightly different to what I had in mind, but not bad.

poke @Wilm0r, requesting opinion about this whole idea

dequis:
Hmm, the diff of this PR still has those protocol-specific `| ACC_SET_LOCKABLE`, aren't those unneeded with the third commit?

Regarding locking `account add`, wouldn't it be desirable to also lock `account del`?

seveas:
I didn't have time to work on these pull requests last week. The protocol-specific lockables are now gone.

And I think the `account del` command doesn't need to be disabled: locked accounts cannot be deleted and I can imagine a setup with locked and unlocked accounts, where being able to delete the latter may be desirable.

conf.c
```diff
@@ -245,6 +246,12 @@ static int conf_loadini(conf_t *conf, char *file)
 			} else if (g_strcasecmp(ini->key, "operpassword") == 0) {
 				g_free(conf->oper_pass);
 				conf->oper_pass = g_strdup(ini->value);
+			} else if (g_strcasecmp(ini->key, "allowaccountadd") == 0) {
+				if (sscanf(ini->value, "%d", &i) != 1) {
+					fprintf(stderr, "Invalid %s value: %s\n", ini->key, ini->value);
+					return 0;
+				}
+				conf->allow_account_add = !!i;
```
dequis: Maybe use `is_bool()` and `bool2int()` here instead of `sscanf` of `%d`?

storage_xml.c
```diff
@@ -79,13 +81,18 @@ static void handle_settings(struct xt_node *node, set_t **head)
 			}
 		}
 		set_setstr(head, name, c->text);
+		if (locked && !g_strcasecmp(locked, "true")) {
+			s = set_find(head, name);
+			if(s)
```
dequis: style nitpick: always use `{...}`. Also, space in `if (s)`

Same thing with the `if(set->flags & SET_LOCKED)` below

dequis:
Other than a couple of nitpicks, this looks good to me.

seveas:
Both comments have been addressed (with no other changes). I didn't know about is_bool/bool2int yet, useful little functions.

dequis:
Thanks!

PR #63, opened 2016-02-23
Support for multiple authentication backends

seveas:
As with PR #61, these patches are driven by my need to operate bitlbee in an
enterprise situation. Integrating with existing authentication solutions is
important there.

This is a 3-patch series that adds two authentication backends and the
infrastructure to easily add more. The infrastructure to do that is inspired by
(but not an exact copy of) the existing storage backend infrastructure.

More details about the changes are in the commits themselves, especially in the
first one.

As both authentication backends require extra libraries and development
headers, they are both optional and bitlbee works as normal without them. Even
with the authentication backends compiled in, they're not used by default and
bitlbee works teh same as now.

### DIFF
--- a/auth_pam.c
+++ b/auth_pam.c
@@ -1,0 +1,62 @@
+#define BITLBEE_CORE
+#include "bitlbee.h"
+#include <security/pam_appl.h>
+
+#define PAM_CHECK(x) do { \
+	ret = (x); \
+	if(ret != PAM_SUCCESS) { \
+		pam_func = #x; \
+		goto pam_error; \
+	} \
+} while(0)
+
+/* This function fills in the password when PAM asks for it */
+int pamconv(int num_msg, const struct pam_message **msg, struct pam_response **resp, void *appdata_ptr) {
+	int i;
+	struct pam_response *rsp = g_new0(struct pam_response, num_msg);
+
+	for (i = 0; i < num_msg; i++) {
+		rsp[i].resp = NULL;
+		rsp[i].resp_retcode = 0;
+		if (msg[i]->msg_style == PAM_PROMPT_ECHO_OFF) {
+			rsp[i].resp = g_strdup((char *)appdata_ptr);
+		}
+	}
+	*resp = rsp;
+	return PAM_SUCCESS;
+}
+
+static storage_status_t pam_check_pass(const char *nick, const char *password)
+{
+	int ret;
+	const struct pam_conv pamc = { pamconv, (void*) password };
+	pam_handle_t *pamh = NULL;
+	char *pam_func;
+
+	PAM_CHECK(pam_start("bitlbee", nick, &pamc, &pamh));
+	PAM_CHECK(pam_authenticate(pamh, PAM_DISALLOW_NULL_AUTHTOK));
+	PAM_CHECK(pam_acct_mgmt(pamh, 0));
+
+	pam_end(pamh, ret);
+	return STORAGE_OK;
+
+pam_error:
+	switch (ret) {
+		case PAM_AUTH_ERR:
+			pam_end(pamh, ret);
+			return STORAGE_INVALID_PASSWORD;
+		case PAM_USER_UNKNOWN:
+		case PAM_PERM_DENIED:
+			pam_end(pamh, ret);
+			return STORAGE_NO_SUCH_USER;
+		default:
+			log_message(LOGLVL_WARNING, "%s failed: %s", pam_func, pam_strerror(pamh, ret));
+			pam_end(pamh, ret);
+			return STORAGE_OTHER_ERROR;
+	}
+}
+
+auth_backend_t auth_pam = {
+	.name = "pam",
+	.check_pass = pam_check_pass,
+};
--- a/root_commands.c
+++ b/root_commands.c
@@ -142,10 +142,9 @@ static void cmd_identify(irc_t *irc, char **cmd)
 		return;
 	}
 
-	if (load) {
+	status = auth_check_pass(irc, irc->user->nick, password);
+	if (load && (status == STORAGE_OK)) {
 		status = storage_load(irc, password);
-	} else {
-		status = storage_check_pass(irc->user->nick, password);
 	}
 
 	switch (status) {
@@ -158,7 +157,6 @@ static void cmd_identify(irc_t *irc, char **cmd)
 	case STORAGE_OK:
 		irc_rootmsg(irc, "Password accepted%s",
 		            load ? ", settings and accounts loaded" : "");
-		irc_setpass(irc, password);
 		irc->status |= USTATUS_IDENTIFIED;
 		irc_umode_set(irc, "+R", 1);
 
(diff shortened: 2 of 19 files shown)


dequis:
Instead of fixing the travis build by installing more dependencies, fix it by making it properly optional in the configure file.

seveas:
It's optional, but enabled by default. And even if it's not enabled by default, I'd like to enable it in the tests so it's tested. There's another build failure with 'make check' that I'm fixing now. I didn't test make check before pushing, I just tested the functionality :smile:

dequis:
In that case please disable it by default and change the ./configure in .travis.yml.

auth_pam.c
```diff
@@ -0,0 +1,62 @@
...
+	if(ret != PAM_SUCCESS) { \
+		pam_func = #x; \
+		goto pam_error; \
+	} \
+} while(0)
+
+/* This function fills in the password when PAM asks for it */
+int pamconv(int num_msg, const struct pam_message **msg, struct pam_response **resp, void *appdata_ptr) {
+	int i;
+	struct pam_response *rsp = (struct pam_response *)g_malloc(num_msg * sizeof(struct pam_response));
```
dequis: `g_new0(struct pam_response, num_msg)`

dequis:
Any idea if this would / should interact in any way with SASL auth? See PR #46, where it says that it's intentionally limited to PLAIN method due to the need to encrypt IM account passwords. That need would be gone, but I guess these two auth backends still require plain passwords.

seveas:
configure/travis now done properly (I hope), and a malloc has been changed to g_new0 per your comment.

I have no idea on the SASL auth interaction I'm afraid. I saw the code, but don't quite know how SASL works so chose to ignore it for now.

seveas:
Actually, the test failure is not because of missing dependencies in the package, but in the unrelated OTR code.

dequis:
```
/usr/include/libotr/auth.h
21:#ifndef __AUTH_H__
22:#define __AUTH_H__
```

The `auth.h` you created breaks it.

seveas:
Heh, long live generic header names...

Good catch, thanks!

seveas:
Just pushed a fresh set of commits as I had left in some pointer signedness warnings

seveas:
This pr and #61 have a small textual conflict. Depending on which one gets merged first, I will rebase the other one and solve the conflict.

seveas:
I'll be changing the first commit in this series significantly to decouple loading rom storage and password checking, please don't merge yet.

I found out later (early this week) that this decoupling is needed to avoid duplicating/triplicating some logic when we have multiple storage backends and certain plugins.

seveas:
Changed commits now pushed. The big difference is that auth_check_pass now calls storage_check_pass and not the other way around.  This lets the auto-registration of users work and also helps some of the PR's to come.

The textual conflict with #61 is still there, I'll rebase whichever gets merged last of these two.

seveas:
Now rebased on top of current master (after the merge of #61) with some style fixes (`if`'s without `{...}`)

storage_xml.c
```diff
@@ -120,29 +118,40 @@ static xt_status handle_account(struct xt_node *node, gpointer data)
...
-			acc->flags |= ACC_FLAG_LOCAL;
-		}
-		if (locked && !g_strcasecmp(locked, "true")) {
-			acc->flags |= ACC_FLAG_LOCKED;
-		}
-	} else {
+	}
+
+	pass_len = base64_decode(pass_b64, (unsigned char **) &pass_cr);
+	if (pass_len < 0) {
```
dequis: I don't think this is ever hit.

root_commands.c
```diff
@@ -267,7 +264,10 @@ static void cmd_drop(irc_t *irc, char **cmd)
 {
 	storage_status_t status;
 
-	status = storage_remove(irc->user->nick, cmd[1]);
+	status = auth_check_pass(irc, irc->user->nick, cmd[1]);
+	if (status == STORAGE_OK)
```
dequis: nit: {...}

root_commands.c
```diff
@@ -142,11 +142,9 @@ static void cmd_identify(irc_t *irc, char **cmd)
 		return;
 	}
 
-	if (load) {
+	status = auth_check_pass(irc, irc->user->nick, password);
+	if (load && (status == STORAGE_OK))
```
dequis: nit: {...}

storage_xml.c
```diff
@@ -239,24 +248,26 @@ static storage_status_t xml_load_real(irc_t *irc, const char *my_nick, const cha
...
 	}
 
-	{
+	if (action == XML_PASS_CHECK) {
 		char *nick = xt_find_attr(node, "nick");
 		char *pass = xt_find_attr(node, "password");
+		char *backend = xt_find_attr(node, "auth_backend");
 
-		if (!nick || !pass) {
+		if (!nick || !(pass || backend))
```
dequis: nit: {...}

dequis:
Nitpicks aside, first commit looks good.

storage_xml.c
seveas: It is hit when feeding it corrupt base64 data, isn't it?

root_commands.c
seveas: Dammit, thought I cauht them all!

storage_xml.c
dequis: Checked glib just to be sure, it returns 0 and a possibly empty allocated string. You could probably remove the whole condition, or switch it to `==`. Whichever you prefer.

dequis:
I'm sorry about ldap forcing you to use deprecated functions (which aren't even mentioned as deprecated in the man page)

I don't really understand most of that code so I'm going to blindly trust you there.

The pam api is slightly weird but still straightforward and easy to understand, but i'm not going to pretend I can say anything meaningful about it.

One thing that caught my attention is that ldap seems to provide some asynchronous lookup api. BitlBee is supposed to be all async nonblocking webscale html9 responsive boilerstrap [add more buzzwords here], but having multi-user installs in daemon mode (as opposed to forkdaemon) isn't a great idea in pratice, so I think it's reasonable to do a couple of blocking calls here and there. And it would suck to have to redesign identify password checking code to be async. Async code sucks in general.

tl;dr looks good.

seveas:
I could use the async lookup API, but that would mean reengineering the whole authentication thing to be async. Painful, and I wouldn't even know how/where to start. So I'm glad you're not forcing me to :smile:

A quick explanation of ldap:

- bind/unbind is login/logout
- you have to bind with a DN, but you only have a loginname, so first we do an anonymous bind (read-only login without user/pass) to map the login (e.g. Seveas) to the DN (e.g. uid=Seveas,ou=People,dc=example,dc=com)
- To do that search, you have to work with ldap filters and messages, which are sql queries and results, except more painful

And PAM:

`pam_authenticate` starts a "conversation/dialog" protocol. As soon as you call pam_authenticate, its internal state mechanism takes over. In theory this can also do 2fa and other fancy things by continuously prompting for more input by calling the conversation function. Our conversation function however only has a password and feeds it when pam asks. If you try anything fancier, we just signal to pam that we don't support it and pam will fail to authenticate the user.
`pam_acct_mgmt` is needed so things like pam_access work.

I'll fix the remaining issues later today, thanks for the review!

seveas:
Fresh commits pushed, the only changes are fixes for the if {...} and base64_decode issues.

dequis:
Thanks!

PR #64, opened 2016-02-23
Un-deprecate the storage_{load,save,remove} plugin hooks

seveas:
They are useful points to do automated account maintenance and policy
enforcement.

I currently have a mysql backend which does unspeakable things to accounts that
it needs to stop doing before I can submit the mysql backend itself as a patch.
For it to stop doing that, I want to write a plugin that does all the dirtty
work. The storage_load hook is exactly the right place to do so.


dequis:
Uhm, while I agree with the change, there's no real need to have this as a separate PR, it has no code changes at all. Do it as part of your mysql thing.

PR #66, opened 2016-02-28
Password modernization

seveas:
I think it's time bitlbee switches away from obsolete password hashes and
encryption. I propose scrypt and aes-gcm to replace them and have implemented
the necessary bits to do so in a backwards compatible way. More details in the
commit messages. 

This is built on top of my authentication backends branch as it would generate
lots of textual conflicts later otherwise, so please ignore the first three
commits. I'm submitting this more to get feedback than to merge in its current
state and will rebase when the authentication backends branch is merged.

The fourth commit is preparatory for both the final commit of this series and
the upcoming mysql backend (which will be built on top of this branch
initially). It centralizes all password handling using generic function names.

The fifth and final commit actually implements the new hashing and encryption
schemes and backwards compatibility code. I've verified that all three
implementations are 100% compatible with each other and with configfiles/xml
files generated by older bitlbees.

<rant>
This required me to dive into the sources of gnutls, nss and openssl as the
documentation is very, very shoddy. OpenSSL's documentation is best, nss the
absolute worst and I could only implement this by reading the guts of the nss
sources. I need a stiff drink now.
</rant>


seveas:
One quite imporant thing to note (and the reason the tests fail on travis, but pass locally for me) is that libscrypt is quite new and may not be widely available. I'm not quite certain whether the added dependency is worth it or not. I could probably implement a pbkdf2 hasher on top of nss/gnutls/openssl instead, or even ship the needed bits of libscrypt with bitlbee (which its license allows). Feedback would be appreciated.

seveas:
Hmm, and even with libscrypt vendored in (28c1ce780d8675ae49200cc1bddec0d93ab54e24 and 90f4670a166e22e93d92096100eea50aefbbcbca), [tests fail on travis' ancient (4 year old) containers](https://travis-ci.org/seveas/bitlbee/builds/112393414) as its gnutls is too old to understand AES-GCM.

dequis:
Oh hey i was looking into doing the same thing a couple of weeks ago. If it was for scrypt only i would have used libgcrypt, which already is sort of a dependency of bitlbee (either through gnutls or through libotr).

However I don't think scrypt is the best idea - its strength depends on its memory hardness, which makes it perfect for things like full disk encryption (done very early in the boot process when you can as much memory as you want) but terrible for a network server like bitlbee. We don't want one password check taking hundreds of mbs of ram.

So my choice was bcrypt, which uses roughly 64kb and is still pretty damn good.

The preferred way to include it seems to be vendoring crypt_blowfish, which is public domain. http://www.openwall.com/crypt/

dequis:
Oh crap you went all the way and implemented the AES-GCM through each ssl lib. Maybe I should have made it more explicit that we sorta depend on libgcrypt already.

seveas:
Thanks for the review!

bcrypt works too, will change to do bcrypt instead of scrypt. Initially I just went for what was considered best of breed, which seems to be scrypt with bcrypt as close second.

About libgcrypt: I was planning to reduce the dependency on libgcrypt by reimplementing the only thing that uses it (apart from the OTR plugin) with native gnutls methods. gnutls doesn't use gcrypt but nettle as crypto backend these days, so seeing gcrypt being used in ssl_gnutls.c felt "wrong" :smile:

So what would you like me to do here? I see 2 possibilities:

- Reimplement ssl_des3_encrypt to not require gcrypt and confine gcrypt usage to the OTR plugin. Leave my current aes functions in place
- Remove ssl_des3_encrypt from ssl_*.c and move the ssl_gnutls.c version to a common file so we don't have 3 copies, and reimplement my aes functions with gcrypt.

I slightly prefer option one, but that's probably only because I already dug into the deep nastinesses of the ssl libraries.

dequis:
Fourth commit (the one that prepares by moving functions around) looks good. You might want to add a comment to password_encode/decode saying what they are for, since without that context they look wrong.

seveas:
bcrypt doesn't provide a hash long enough to use as AES key, so can't be used to replace scrypt for that purpose. gcrypt does support pbkdf2, which seems only slightly worse than bcrypt. This makes me think I should go for option 2 and [gcrypt all the things](http://i.imgur.com/8N2I4Lr.jpg).

dequis:
re: KDF: I had a long paragraph here mentioning that maaaybe you could adapt bcrypt and maaaaybe you could use scrypt with 16mb ram, but honestly it doesn't matter that much when we're currently using MD5. And we've stayed with MD5 for this long mostly because it doesn't matter that much. So i'd say just use pbkdf2

re: AES, I never got to consider the IM account passwords part too much because I don't think I understand enough about symmetric encryption to make a safe decision there. Of course it was silly to pretend that it's safe to leave it with rc4, especially when I'm aware of a number of attacks that AES-GCM is good against but I'm still worried about the possibility of attacks I'm not aware of. Meanwhile the rc4 implementation... okay I don't want to think about that one too much.

Anyway AES-GCM is a pretty good option as far as I know.

Another other option I was considering was nacl's `crypto_secretbox`, possibly by embedding tweetnacl. The advantage would be that the nacl api is designed to be explicit about everything. Sorta.

Improper api usage is likely to be the main concern with this stuff, which is why I really don't want three implementations.

One thing that has been bothering me for a while - if you use a KDF to hash the password and the same output as the AES key, wouldn't that mean the key for the passwords is saved in "plaintext" in `<user ... password="hash here" ...>`?

seveas:
Ok, pbkdf2 and gcrypt it is. I share your concern about API misuse, though by having to implement it 3 times with wildly varying API's, I think I did catch all the API misuse I could. So I'll implement it with gcrypt and verify the results with the 3 other APIs to make sure all matches.

dequis:
btw you're great

seveas:
OMG is gcrypt a nightmare. I can't make it produce the same output as openssl/nss/gnutls for AES256/GCM. For starters, it enforces a tag size of 16. But even when modifying the gnutls impleementation to have 16 byte tags, the outputs are not compatible. When hardcoding a fixed IV, the outputs differ between gcrypt and gnutls! And not just the tag, the encrypted data already.

Since gnutls, openssl and nss all give the same output, I'm almost inclined to believe gcrypt is buggy. But I'll keep looking for a bug in my code.

seveas:
pbkdf2 and aes-gcm-via-gcrypt it has become. Tests fail on travis, the ancient gcrypt it has is too old to support aes-cgm. Tests succeed locally.

seveas:
Added two commits and tests now work on travis by switching to trusty and recent otr/gcrypt.

dequis:
Ugh that's annoying. Debian wheezy (oldstable) and ubuntu precise 12.04 LTS should be kept working - they are still supported and widely used. But both scrypt and aes-gcm were introduced in libgcrypt 1.6.0 and they have 1.5.0.

Maybe have a configure parameter to opt-in to use the rc4/md5 crap? It sucks but we can get rid of that once those distros are unsupported.

This doesn't really help with the plan of making libgcrypt a requirement for all installs...

seveas:
Yeah, if you want to keep supporting those old releases, old gcrypt is the way to go. I'll put back the rc4/md5 writing code and autodetect gcrypt support for good ciphers.

seveas:
I've put back the arc4 code and done some preprocessor things to use it when gcrypt is too old. old-gcrypt does support pbkdf2, so I haven't revived the md5sum writing code.

Tests now pass on travis without modifications to .travis.yml, so once the authentication backends PR is in, I'll rebase this once more to make it ready.

seveas:
I heard @dequis was getting bored, so I rebased this on top of current master, ready for review and maybe even merge :see_no_evil:

lib/password.c
```diff
@@ -1,32 +1,36 @@
...
-
-	/* Save the hash in base64-encoded form. */
-	*hash = base64_encode(pass_md5, 21);
+	random_bytes(salt, saltlen);
+	gcry_kdf_derive(password, strlen(password), GCRY_KDF_PBKDF2, GCRY_MD_SHA1,
+			salt, saltlen, rounds, keylen, rawhash);
+	b64salt = base64_encode(salt, saltlen);
+	b64hash = base64_encode(rawhash, keylen);
+
+	*hash = g_strdup_printf("$1$%s$%d$%s", b64salt, rounds, b64hash);
```
dequis: `$1$` means md5_crypt http://pythonhosted.org/passlib/modular_crypt_format.html#mcf-identifiers

As the above page explains there's no standard for this stuff but they seem to use `$pbkdf2$` for this case, with rounds first and salt second.

lib/password.c
```diff
@@ -1,32 +1,36 @@
...
-	random_bytes(pass_md5 + 16, 5);
-	md5_init(&md5_state);
-	md5_append(&md5_state, (md5_byte_t *) password, strlen(password));
-	md5_append(&md5_state, pass_md5 + 16, 5);   /* Add the salt. */
-	md5_finish(&md5_state, pass_md5);
-
-	/* Save the hash in base64-encoded form. */
-	*hash = base64_encode(pass_md5, 21);
+	random_bytes(salt, saltlen);
+	gcry_kdf_derive(password, strlen(password), GCRY_KDF_PBKDF2, GCRY_MD_SHA1,
```
dequis: Maybe `GCRY_MD_SHA256` instead of `GCRY_MD_SHA1`?

tests/check_arc.c
```diff
@@ -99,7 +69,6 @@ Suite *arc_suite(void)
 	TCase *tc_core = tcase_create("Core");
 
 	suite_add_tcase(s, tc_core);
-	tcase_add_test(tc_core, check_codec);
```
dequis: Maybe keep these tests? Not the most important thing but no reason to nuke them

lib/password.c
```diff
@@ -0,0 +1,237 @@
...
+/* pbkdf2-hashed password. We store an algoritm identifier ("1"), the salt
+ * used, the number of rounds and the final hash */
+size_t password_hash(const char *password, char **hash)
+{
+	unsigned char rawhash[64], salt[8];
+	char *b64hash, *b64salt;
+	int saltlen = 8, rounds = 10000, keylen = 64;
+
+	random_bytes(salt, saltlen);
+	gcry_kdf_derive(password, strlen(password), GCRY_KDF_PBKDF2, GCRY_MD_SHA1,
```
dequis: Maybe `GCRY_MD_SHA256` instead of `GCRY_MD_SHA1`?

lib/password.c
seveas: In absence of a standard, a good example will do, and passlib is as good as any, will change.

tests/check_arc.c
seveas: ok.

lib/password.c
seveas: From what I've read, doing sha1 is standard here. I know sha256 is stronger but the crypto community seemed ambivalent about it. I don't care much either way, so I'll change it.

dequis:
I've poked other people to take a look at the crypto here, and they may or may not comment here. Because I don't feel qualified to give this a complete approval, even though it's obvious it's a huge improvement over the previous code.

@seveas You have a small merge conflict btw, not sure when that happened.

seveas:
@dequis did you get anywhere with reviews in the last 3 months?

PR #67, opened 2016-02-28
Mysql storage backend [do not merge in current form]

seveas:
As threatened, here is a MySQL storage backend for BitlBee. As it depends on my
locked down accounts, authentication backend and password modernization
branches, it is absolutely not suitable as is for merging, consider this a preview.

It does work though, quite well even :smile:

All but the last three commits are part of other PR's or merges with them,
please ignore those. The third to last one fixes a minor annoyance around early
logging that got in the way of debugging. The second to last fixes a bug in the
handling of non-default storage backends and the last bug is the big commit
with the new backend.

It's completely optional and configure will not even complain if you don't have
the mysql headers installed. But if you do have them installed, it will build
the backend. I could add a --mysql=[01] knob if desired, but autodetecion works
for me.

I've added the schema in doc/ and updated the example config to mention the
AccountStorage setting, but have not added it to the documentation yet.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 language: c
 
 script:
- - ./configure --pam=1
+ - ./configure --pam=1 --ldap=1
  - make check
  - BITLBEE_SKYPE=plugin dpkg-buildpackage -uc -us -d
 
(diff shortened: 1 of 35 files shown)


seveas:
[Note to self: the auto_connect logic is inverted, fixed in the policy-plugin branch, needs squashing before resubmitting]

PR #68, opened 2016-02-29
away-notify: Send new away message when message changes

digitalcircuit:
Send new away message when ```away-notify``` is enabled and the status changes (e.g. "Away" to "Mobile") or the status message changes without returning from away.

This fixes IRC clients with ```away-notify``` support persisting old away messages.


dequis:
Freeze is over, so merged this one as f892236adc18544c0ff424a18e21f98c5962ce12. Thanks!

PR #71, opened 2016-03-26
Fix use of un-initialized variable

mhalden:
On line [127 of storage_xml.c](https://github.com/bitlbee/bitlbee/blob/8e6ecfe23ff985e57675bd00b94860edb62de9ad/storage_xml.c#L127) there is a use of `pass_len` which has not yet been initialized. This made bitlbee return "Unknown error while loading configuration" on login. This PR fixes this by correctly initializing the variable before it's used.


dequis:
Yay for last minute changes that break things. Thanks

PR #72, opened 2016-04-06
Fix segfault when CAP END is received multiple times

mhalden:
If bitlbee receives the command CAP END after the user is registered it will currently segfault. This PR fixes this by ignoring CAP END after the user is registered (which is according to the spec is the correct behavior).


dequis:
Well, fuck. Does this happen in any real world situations?

mhalden:
None that I know of, I triggered this by manually sending CAP END when I was looking into how capabilities work. The only other possibility I can think of is if a client is not behaving correctly.

dequis:
Cool then. Luckily this isn't as critical as it might seem because most servers run in forkdaemon mode, which means the damage is limited to a single client connection.

Thanks for the fix!

PR #73, opened 2016-05-14
Twitter: add "like" as an alias for favourite.

AaronVanGeffen:
This will allow a user to type "like" in Twitter channels, reflecting recent changes to Twitter itself. Note that the API hasn't changed.


dequis:
This should probably be mentioned in the documentation, but it doesn't mention the other aliases either... https://github.com/bitlbee/bitlbee/blob/master/doc/user-guide/commands.xml#L887

AaronVanGeffen:
Do you want me to append them to the `listitem` tag on that line, or would you rather I add them to `term` itself?

dequis:
Merging this to get it out of the way, going to take care of the twitter docs later. "help set commands" is big enough to split it to a different help topic, so i'd like to reorganize things there.

PR #74, opened 2016-05-15
Remove the ARCH / CPU defines.

jelmer:
These only reflect on what arch/cpu bitlbee was built, not on which
it is running. This makes the Debian package unreproducible.
See e.g. https://tests.reproducible-builds.org/rb-pkg/testing/i386/bitlbee.html

### DIFF
--- a/configure
+++ b/configure
@@ -56,9 +56,6 @@ ldap=0
 
 pie=1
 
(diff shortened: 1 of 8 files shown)


dequis:
Aww, i will miss them :(

But okay.

PR #76, opened 2016-05-15
Implemented plugin information for external plugins

jgeboski:
As of now, bitlbee will load any plugin regardless of the ABI it was
built against. This is really problematic when structures or symbols
are changed within bitlbee. This often leads to the plugin not loading
or the plugin acting in an undefined way. Typically a simple rebuild of
the plugin will resolve such issues, but many users have no idea that
this is required after they have updated bitlbee.

Furthermore, it is often times impossible to determine the version of
a plugin, without relying on the package manager of the system. This is
quite a problem when users are reporting bugs for external plugins, and
they have no idea what version of the plugin they are running. This is
also an opportunity to provide additional metadata for each plugin that
can then be displayed to the user.

Solving these issues is done by adding a new required function to each
plugin. The init_plugin_info() function must now be implemented along
with the init_plugin() function. This function then returns a static
structure, which retains all of the metadata for the plugin. Then this
is used by bitlbee to check the ABI version and provide information to
the user.

The introduction of the new function is required as bitlbee needs to
obtain the ABI version before calling init_plugin().

The boiler-plate implementation of init_plugin_info():

```
#ifdef BITLBEE_ABI_VERSION_CODE
struct plugin_info *init_plugin_info(void)
{
    static struct plugin_info info = {
        BITLBEE_ABI_VERSION_CODE,            /* Required */
        "plugin-name",                       /* Required */
        "1.3.3.7",                           /* Required */
        "A short description of the plugin", /* Optional */
        "First Last <alias@domain.tld>",     /* Optional */
        "http://www.domain.tld"              /* Optional */
    };

    return &info;
}
#endif
```

The example wraps the function declaration in an if block for backwards
compatibility with older bitlbee versions.

Displaying the plugin metadata is done via the newly added "plugins"
command, which simply dumps formatted data to the root channel.

### DIFF
--- a/bitlbee.h
+++ b/bitlbee.h
@@ -39,6 +39,7 @@ extern "C" {
 #define VERSION BITLBEE_VERSION
 #define BITLBEE_VER(a, b, c) (((a) << 16) + ((b) << 8) + (c))
 #define BITLBEE_VERSION_CODE BITLBEE_VER(3, 4, 2)
+#define BITLBEE_ABI_VERSION_CODE 1
 
 #define MAX_STRING 511
 
--- a/doc/user-guide/commands.xml
+++ b/doc/user-guide/commands.xml
@@ -1835,12 +1835,12 @@
 	</bitlbee-command>
 
 	<bitlbee-command name="plugins">
-		<short-description>List all the external plugins</short-description>
+		<short-description>List all the external plugins and protocols</short-description>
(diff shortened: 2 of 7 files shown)


dequis:
@sm00th relevant to your interests!

dequis:
ty

PR #78, opened 2016-06-10
Use the correct error message when no authentication schemes are supported

mhalden:
This will make bitlbee tell the user about the requirement for oauth when the server actually announces support for oauth. If the server does not announce oauth support bitlbee will tell the user it doesn't support any of the schemes provided by the server.

These messages were reversed before.


dequis:
Thanks!

PR #80, opened 2016-06-26
Added an interface for the listing of existing chatrooms

jgeboski:
Several protocols can provide a list of existing chatrooms that a user
is able join. This is crucial for the usage of several protocols, most
notably Purple and Facebook.

Plugins wishing to support this extended functionality must implement
the new prpl->chat_list() function. This implemented function will in
most cases send a remote request for the list of chatrooms. Once the
list of chatrooms is obtained, a bee_chat_info_t GSList must be created
and assigned to the im_connection->chatlist field. Then a call to the
bee_chat_list_finish() is needed to display the list to the user.

The chat list is maintained entirely by the plugin, so it is important
to ensure all pointers related to the chat list remain valid until the
chat list is set to NULL. This list is used internally by bitlbee to
calculate indexes, which then allows the user to join a chat with an
index, rather than some random identifier. It also important to ensure
the list is properly freed whenever it is updated, or when the account
is disconnect via the prpl->logout() function.

On the user interface side of things, the 'chat list' subcommand was
recommissioned. For a user to list the existing chat rooms:

    chat list <account id>

Afterwards a user can join a chatroom in the list with its index. This
extends the functionality of the 'chat add' subcommand by adding in
support for the exclamation point operator to denote an index.

    chat add <account id> !<index> [channel]

### DIFF
--- a/bitlbee.h
+++ b/bitlbee.h
@@ -172,6 +172,7 @@ void root_command_string(irc_t *irc, char *command);
 void root_command(irc_t *irc, char *command[]);
 gboolean root_command_add(const char *command, int params, void (*func)(irc_t *, char **args), int flags);
 gboolean cmd_identify_finish(gpointer data, gint fd, b_input_condition cond);
+void cmd_chat_list_finish(struct im_connection *ic);
 gboolean bitlbee_shutdown(gpointer data, gint fd, b_input_condition cond);
 
 char *set_eval_root_nick(set_t *set, char *new_nick);
--- a/doc/user-guide/commands.xml
+++ b/doc/user-guide/commands.xml
@@ -263,7 +263,7 @@
 		</description>
 
 		<bitlbee-command name="add">
-			<syntax>chat add &lt;account id&gt; &lt;room&gt; [&lt;channel&gt;]</syntax>
+			<syntax>chat add &lt;account id&gt; &lt;room|!index&gt; [&lt;channel&gt;]</syntax>
 
 			<description>
 				<para>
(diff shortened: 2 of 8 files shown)


jgeboski:
Couple things @dequis:

1. Pretty sure we need to bump the ABI version since we are modifying the `im_connection` structure, which has no padding. I know you wanted to avoid bumping.
2. The purple stuff is going to get a bit weird with the server listing, see Jabber. The problem is that libpurple does not have a concise way for us to get/set the server, as it relies on its request API. There is a `server` field with Jabber rooms, but other protocols might not do this. We could have the user type the server twice, once for the `chat list` command, and once for the libpurple request, but that is just lame.

dequis:
1. I'd hate to have to require a specific abi version already. I don't think we've made the missing abi version errors noticeable enough so that people will realize what's going on once we make it required. You can, and should, bump the version, but not sure about requiring it.

    I think it's still safe as long as you don't try to access that member of `im_connection` if `prpl->chat_list`  wasn't implemented

2. It's unfortunate but I think it's acceptable to ignore this issue.

My main concern with this thing is the ownership of purple roomlist, that stuff wasn't clear at all when I looked at it.

jgeboski:
Okay, I have added an additional check so that an ABI version bump is not needed.

mhalden:
There seems to be an issue with `chat list <offline account>` leading to segfault. I think the problem is that `ic` may be `NULL` when the account is not logged in. Because of this there is a null-pointer dereference in `purple_chat_list` when the account is not signed in.

I'm not sure it's the best way to do it, but [this patch](https://lden.im/~marius/files/bitlbee-chatlist.diff) seems to work for me.

dequis:
Neat, thanks for having a look!

jgeboski:
@mhalden good catch! Your patch looks good. Would you mind providing a Git formatted patch? Or should I just attribute it to you with a commit message of my own?

mhalden:
I have updated [the patch](https://lden.im/~marius/files/bitlbee-chatlist.diff) to be git formatted.

jgeboski:
@mhalden Thanks for testing and patching!

@dequis This branch has been rebased to `master`.

mumixam:
line 260 in commands.xml needs `list` added

dequis:
@mumixam I applied your suggested changes and a couple of others as 028ca92f3a36fed8ebb0a6507e64a50938049e53, thanks

PR #82, opened 2016-09-20
twitter: Changes to 140 char limit detection

WPettersson:
As of the 19th of September, Twitter will apparently no longer consider
@usernames at the start of a tweet as part of the 140 characters. URLs are
already ignored when determining this limit, so this patch uses a similar
approach to not counting @usernames.

It's important to note that these are only ignored if at the start of the
tweet, but there's no hard source from Twitter on how this is calculated. For
now, I'm assuming that "@username" at the start is skipping, as is
"@username1 @username2" and so-on.

I don't think this should be merged in until we have a definite source on what is or is not counted, but at least this is a start to adopting this change.

Source: http://www.theverge.com/2016/9/12/12891562/twitter-tweets-140-characters-expand-photos

### DIFF
--- a/protocols/twitter/twitter.c
+++ b/protocols/twitter/twitter.c
@@ -462,6 +462,7 @@ int twitter_url_len_diff(gchar *msg, unsigned int target_len)
 	int url_len_diff = 0;
 
 	static GRegex *regex = NULL;
+	static GRegex *name_regex = NULL;
 	GMatchInfo *match_info;
 
 	if (regex == NULL) {
@@ -477,6 +478,21 @@ int twitter_url_len_diff(gchar *msg, unsigned int target_len)
 
 		g_free(url);
 		g_match_info_next(match_info, NULL);
+	}
+	g_match_info_free(match_info);
+
+	if (name_regex == NULL) {
+		name_regex = g_regex_new("^@\\w+( @\\w+)?", 0, 0, NULL);
+	}
+
+	g_regex_match(name_regex, msg, 0, &match_info);
+	if (g_match_info_matches(match_info)) {
+		gchar *url;
+
+		url = g_match_info_fetch(match_info, 0);
+		url_len_diff += target_len - g_utf8_strlen(url, -1);
+
+		g_free(url);
 	}
 	g_match_info_free(match_info);
 


dequis:
Fancy, thanks for the PR.

There are better docs over here: https://dev.twitter.com/overview/api/upcoming-changes-to-tweets

The docs mostly talk about consuming, but I believe it's trying to say that the hidden prefix is only going to be the auto-populated one.

WPettersson:
I spent 30 minutes trying to find that source, but couldn't. Probably because I was expecting a mention of "September 19".

Ok, that changes things. In particular, it does mean this fix is plain wrong. The right thing to do is to use the new auto_populate_reply_metadata when sending replies (which also means that "reply XX" can more easily be used when in group conversations). That should be an easy start.

The extended_tweets change is something I might look at later, separately. In particular, there's a bit saying that replies in the hidden prefix "shall not be rendered as part of the display text, but must instead be rendered as metadata" which I think we have to plain ignore, despite going against RFC 2119.

dequis:
Yeah, i've been following this for a while, one of the cleanest feature deployments they've done lately IMO (mostly because the rest tend to be messy).

`auto_populate_reply_metadata` looks great, and exactly what we needed to make the reply command look more like the web one without storing tweet bodies / looking them up before replying, which also would have the side effect of reducing the tweet length in a confusing way. I'm pretty happy about this change.

There's also this tag in their forums with more questions, worth reading too: https://twittercommunity.com/tags/extended-tweet

Also, next time please join irc and talk about what you plan to do, especially if it's big stuff.

Wilm0r:
I guess you talked on IRC about this by now or something?

IMHO we should rip out this code and just let Twitter do this and change the counting whenever they feel like it (also think of t.co changes, where I think they're always doing https now and BitlBee is already off by one for non-https links?), as they do often enough. There are dumb corner cases anyway like the username has to exist I think?

So as long as the error coming back from them has an official count, why keep trying to replicate the effort?

WPettersson:
Haven't spoken, was at work, but just logged in to IRC.

I do like the ripping-out idea. There are too many corner cases for us to catch them all, and all we really gain is a few seconds difference between a local bitlbee server saying "too long" and Twitter saying "too long".

Wilm0r:
Yup, that exactly. It'd only be annoying if the error msg from Twitter did not include their count. I don't know whether it does, but even if it did at least so far BitlBee did not bother to show it I think.

WPettersson:
Bitlbee does show something, it says "Maximum message length exceeded: 341 > 140". From what I can tell, though, Twitter will just send back a generic "Tweet is too long" error in a status-403 response.

dgw:
Yes, I tested this in Twitter's API console (against API 1.1). The response is about as bare-bones as it gets.

It's too bad, really. Twitter could be a bit more helpful, one would think, given that they're counting-this-but-not-that these days.

````HTTP
HTTP/1.1 403 Forbidden
x-frame-options: SAMEORIGIN
last-modified: Tue, 20 Sep 2016 09:42:10 GMT
status: 403 Forbidden
Content-Length: 68
x-response-time: 84
Connection: keep-alive
x-transaction: <snipped>
Server: tsa_b
pragma: no-cache
cache-control: no-cache, no-store, must-revalidate, pre-check=0, post-check=0
x-connection-hash: <snipped>
x-xss-protection: 1; mode=block
x-content-type-options: nosniff
expires: Tue, 31 Mar 1981 05:00:00 GMT
Date: Tue, 20 Sep 2016 09:42:10 GMT
set-cookie: lang=en; Path=/
set-cookie: guest_id=v1%3A<snipped>; Domain=.twitter.com; Path=/; Expires=Thu, 20-Sep-2018 09:42:10 UTC
content-disposition: attachment; filename=json.json
x-twitter-response-tags: BouncerCompliant
strict-transport-security: max-age=631138519
x-access-level: read-write-directmessages
Content-Type: application/json;charset=utf-8
````
````JSON
{
  "errors":  [
     {
      "code": 186,
      "message": "Status is over 140 characters."
    }
  ]
}
````

Wilm0r:
Crap. Yeah, I know BitlBee does the reporting by itself but that's the code we really should just kill. But that error msg from Twitter sure is utterly unhelpful indeed, argh. :<

I guess there's a separate API call to get an authoritative count? (Nice race condition! :-P) Really would hate to have to use that as well.

dgw:
Unless I'm completely overlooking it in [the docs](https://dev.twitter.com/rest/public)…no. There's no API call to ask Twitter how long a tweet is, or if the content is too long, without attempting to post it.

Now that I think about it, though, why would they want to make it easier for third-party clients to do their work? We all know that Twitter has been making it harder and harder for API clients over the last few years. (The per-app [token limits](https://dev.twitter.com/overview/terms/policy#7.Update_Avoid_Replicating_the_Core_Twitter), [for example](http://www.androidpolice.com/2016/04/25/fenix-pulled-from-play-store-after-hitting-twitters-dumb-token-limit/).)

PR #84, opened 2016-09-22
purple: support setting chat room topics

EionRobb:


dequis:
ty

dequis:
Changing the topic works, but hangouts doesn't seem to report topic changes to the UI. This applies to pidgin too, so not an issue here, merging.

PR #85, opened 2016-09-24
irc: Use the standard format for away replies 305/306

arza-zara:


dequis:
There's no standard for that part, it's an opaque meaningless string that clients should not parse. Is this an actual issue anywhere?

arza-zara:
It's not an actual issue, it's just how it's in the RFCs and how every IRC server I know displays them.
> 305	RPL_UNAWAY	":You are no longer marked as being away"
> 306	RPL_NOWAWAY	":You have been marked as being away"

dequis:
Well, if it's not broken, don't fix it.

I appreciate the attention to detail but the IRC RFCs are not the right place for that - aside from the fact that they are very incomplete, if you look closely they have plenty of wrong / misleading things (like invalid BNFs). In this case it's not "wrong", just a matter of not taking it as gospel.

PR #86, opened 2016-10-05
misc.c: Return early in strip_html if the input is zero length

mcm:
If `strip_html` is `always`, and the client tries to join an XMPP (and maybe others) groupchat where the topic is blank or empty, `strip_html` calls strcpy to copy a 1-byte array to a 0-byte array. This just returns early in strip_html if `in` is zero-length to avoid the issue altogether.

I'm not sure if there's anything else to provide with this, but let me know if so and I'm happy to do what I can.

### DIFF
--- a/lib/misc.c
+++ b/lib/misc.c
@@ -162,6 +162,13 @@ void strip_html(char *in)
 
 	memset(out, 0, sizeof(out));
 
+	if (strlen(in) == 0) {
+		/* If strlen(in) == 0, there's no point processing it. Also, the strcpy at the end
+		   of this function will EXC_BAD_ACCESS because strlen(out) == 1. So just bail out
+		   now. */
+		return;
+	}
+
 	while (*in) {
 		if (*in == '<' && (g_ascii_isalpha(*(in + 1)) || *(in + 1) == '/')) {
 			/* If in points at a < and in+1 points at a letter or a slash, this is probably


dequis:
Neat

lib/misc.c
```diff
@@ -162,6 +162,13 @@ void strip_html(char *in)
 
 	memset(out, 0, sizeof(out));
 
+	if (strlen(in) == 0) {
```
dequis: You could do `if (!*in)` here so that it's equivalent to the check below - calling strlen isn't really needed anyway.

mcm: That works too. For my own morbid curiosity because C is FAR from my best language - what's the benefit to one over the other?

dequis:
Got a backtrace of the xmpp issue? I suspect you're fixing the wrong thing here. If the input is an empty string, `strcpy(in, "")` should only write one \0 there and succeed. The problem seems to be that it's trying to write to read-only memory.

lib/misc.c
dequis: It's reading the first byte of the string VS calling one function which may iterate all the way to the end of the string to return a length that isn't used. Unlikely to be a big deal, so i'd rather do it for the sake of doing it the same way as the block below.

(That being said, see my other comment, the whole thing might be wrong :D)

mcm:
Closing this since its fixed by 9698fc0a190f958fa28a8e54b2bfd6ccfdb2aa18

PR #87, opened 2016-10-07
Update WhatsApp plugin for yowsup 2.5

stevenjm:
The removed imports appear to be unnecessary -- nothing else uses those
names in this file -- and they have been removed in yowsup. The other
change is due to a change in API in the YowsupEnv module.

I'm not certain whether this commit is sufficient or even correct, but
it has made the plugin work for me.

### DIFF
--- a/python/wa.py
+++ b/python/wa.py
@@ -44,8 +44,6 @@ from yowsup.layers.protocol_privacy.protocolentities     import *
 from yowsup.layers.protocol_profiles.protocolentities    import *
 from yowsup.layers.protocol_receipts.protocolentities    import *
 from yowsup.layers.axolotl.protocolentities.iq_key_get import GetKeysIqProtocolEntity
-from yowsup.layers.axolotl import YowAxolotlLayer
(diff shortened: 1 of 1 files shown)


Wilm0r:
Wow, I'm actually surprised that someone is using this plugin with
success - it stopped working for me back in March and I had no luck
getting it back to work.

You managed to use the yowsup cmdline utilities to register your
account? Every time I use that now the number gets blocked permanently.
:-(

stevenjm:
> You managed to use the yowsup cmdline utilities to register your account?

I did, just following the example here:

https://github.com/tgalal/yowsup/wiki/yowsup-cli-2.0

It requires the very latest S40 env from GitHub; even the last release from May fails with a message about the version being too old.

I only started trying a few hours ago, so it's possible they may yet ban me, but so far it's working and I've been able to send and receive messages via WhatsApp over IRC.

Wilm0r:
That's great to hear, I'll see if it works for me again as well then, and perhaps FINALLY merge this branch (WA breaking kinda took away my motivation).

PR #88, opened 2016-10-11
purple: allow libpurple message processing before passing to bitlbee.

event:
Remove write_im and write_chat callbacks. Do all handling in write_conv.
This way message is passed through standard libpurple processing (incl. logging) prior to bitlbee handling

### DIFF
--- a/protocols/purple/purple.c
+++ b/protocols/purple/purple.c
@@ -1027,14 +1027,18 @@ void prplcb_conv_del_users(PurpleConversation *conv, GList *cbuddies)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


dequis:
>standard libpurple processing (incl. logging)

Is the purpose of this change to enable logging, or what?

Seems odd that implementing uiops would have that effect, i'll have a deeper look at this later.

dequis:
Did you test this? Every message i send in both private and groupchats gets me the same message echoed back

event:
I did, but apparently not good enough. I'll have this fixed and place another request.

dequis:
You can update pull requests by pushing to the same branch, no need to open a new one

I still would like to know why you're doing this.

PR #89, opened 2016-10-11
install script done

TheTesla:
After recognizing that gnutls version is compiling and running without error messages but is NOT able to make a working tls connection, I wrote this script to install openssl version of bitlbee with all dependencies. BTW the official bitlbee package in ubuntu 14.04 and debian wheezy are also broken. After jabber server answers to proceed tls connection, bitlbee does not answer and says only "Login error: Could not connect to server". Use this install script to avoid this error.

### DIFF
--- a/install.bash
+++ b/install.bash
@@ -1,5 +1,5 @@
 #!/bin/bash
 apt update
-apt install libssl-dev
+apt install libssl-dev pkg-config libglib2.0-dev
 ./configure --ssl=openssl && make && make install && make install-etc && make install-dev
 
--- a/install.bash
+++ b/install.bash
@@ -1,0 +1,5 @@
+#!/bin/bash
+apt update
+apt install libssl-dev
+./configure --ssl=openssl && make && make install && make install-etc && make install-dev
+


dequis:
If you have problems with gnutls, it would be helpful to try to figure out what those problems are, since like you said that's what most distros use and it's the most supported ssl module. OpenSSL is not recommended as the warnings ./configure spews should have told you.

Please join #bitlbee in irc.oftc.net, let's talk about this and get it fixed the right way.

dequis:
The issue turned out to be this: https://wiki.bitlbee.org/FAQ#Can.27t_connect_to_servers_that_use_cacert_certificates_such_as_jabber.ccc.de

>The error looks like this:
> 
>     <@root> jabber - Logging in: Converting stream to TLS
>     <@root> jabber - Login error: Could not connect to server
> 
> It affects the following ubuntu/debian releases (Newer releases are not affected)
> 
>  * Debian 7 (wheezy, oldstable)
>  * Ubuntu 12.04 (precise)
>  * Ubuntu 14.04 (trusty)
> 
> Solutions: (pick one)
> 
>  * Ubuntu 14.04 users can fix this by using the APT repo in http://code.bitlbee.org/debian/ which builds against gnutls28
>  * Debian 7 and ubuntu 12.04 users can upgrade to a newer debian/ubuntu release
>  * Install gnutls28 (possibly in backports) and rebuild bitlbee against it
>  * Rebuild bitlbee with `./configure --ssl=nss` to not use gnutls at all.
>  * Not recommended, but downgrading your gnutls26 package to not include the CVE-2015-7575 fix will also allow you to connect again.
> 
> Context: The ubuntu/debian gnutls26 package [introduced a security "fix"](https://bugs.launchpad.net/ubuntu/+source/gnutls26/+bug/1553819) that prevents connections to servers with certificates with RSA-MD5 signatures. Cacert has that kind of signatures, and jabber.ccc.de is the most common server that uses those certificates.
> 
> The real solution here is for cacert to stop sucking ([they are working on it](http://blog.cacert.org/2016/03/successful-root-re-sign/)... slowly), or to have jabber.ccc.de use a real CA like letsencrypt. Both are out of our control.

PR #90, opened 2016-10-14
purple: allow libpurple message processing before passing to bitlbee.

event:
I've made this mostly for logging. However it generally allows a message to pass through purple processing which at the moment  also emits a purple dbus signal.

### DIFF
--- a/protocols/purple/purple.c
+++ b/protocols/purple/purple.c
@@ -1027,46 +1027,31 @@ void prplcb_conv_del_users(PurpleConversation *conv, GList *cbuddies)
 	}
 }
 
-/* Generic handler for IM or chat messages, covers write_chat, write_im and write_conv */
-static void handle_conv_msg(PurpleConversation *conv, const char *who, const char *message_, guint32 bee_flags, time_t mtime)
+/* Handles write_conv. Passes messages received by purple further up to bitlbee
+ * There are more events which might be handled in the future, but some are tricky.
+ * (images look like <img id="123">, what do i do with that?) */
+static void prplcb_conv_write(PurpleConversation *conv, const char *who, const char *alias, const char *message_,
+                              PurpleMessageFlags flags, time_t mtime)
 {
+	if (flags & PURPLE_MESSAGE_SEND) {
+		return;
+	}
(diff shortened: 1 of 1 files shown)


dequis:
This removes OPT_SELFMESSAGE support...

event:
I thought it's not required as it's shown already, no?
Is there a test suite you could recommend? I feel bad about spending your time like that.

dequis:
There's no (useful) test suite. In this case i'd just connect some purple protocol, send/receive a PM, send/receive a groupchat message, and send a PM and a groupchat message from a different client connected to the same account with one of the libpurple protocols [listed here][1].

The expected behavior is no echoes for sent messages, private self-messages being shown and groupchat self-messages being skipped since it's not possible to tell them apart from echoes as ba7618d607898050db6e7c380e70c77116100d5f says.

Handling self-messages properly is not obvious so please read and understand that commit and the tradeoffs involved. In particular, your code seems to be doing the same thing as the implementation before that commit regarding the PURPLE_MESSAGE_SEND flag

```c
	/* ..._SEND means it's an outgoing message, no need to echo those. */
	if (flags & PURPLE_MESSAGE_SEND) {
		return;
	}
```

Which is fine but sometimes "outgoing message" covers things that were just written locally, things are were echoed by the server asynchronously after being sent by our client (notably in jabber groupchats), and things that were sent by other clients connected to the same account and we must try to show whenever posssible by setting the OPT_SELFMESSAGE flag (which is always 0 in your version)

That aside, I have to say I don't like the idea of enabling purple-side logging. Bitlbee doesn't log chats and it would be a bad idea to start doing so only for the purple protocols. I'd really recommend having your irc client handle logging, or if you use several clients, ZNC. I might be ok with it if logging is turned off by default.

Maybe consider some other method to have the purple built-in message handler called, which doesn't involve giving up the handling of a couple of uiops?

[1]: https://wiki.bitlbee.org/SelfMessages

event:
Ok, I see your point. It seams too complicated and breaking general bitlbee logging approach. It would also enable logging by default which could surprise someone in a bad way. I'll drop this altogether and use client side logging.
Thanks.

PR #91, opened 2016-12-26
[debian] Replace netstat by ss

jspricke:
Also replace the dependency on net-tools by iproute2.
See: https://lists.debian.org/debian-devel/2016/12/msg00604.html

### DIFF
--- a/debian/bitlbee-common.config
+++ b/debian/bitlbee-common.config
@@ -12,13 +12,7 @@ if [ -n "$BITLBEE_PORT" ]; then
 else
 	db_get bitlbee/serveport
 	if [ "$RET" = "stillhavetoask" ]; then
-		listens=$(netstat -ltn | awk '{print $4}')
-		for port in 6667 6666 6668 6669; do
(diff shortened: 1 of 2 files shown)


jelmer:
Is there something we could use that would also work on e.g. kfreebsd-{amd64,i386}? netstat is at least provided by both net-tools and freebsd-net-tools.

jspricke:
Good point, `sockstat -l` should work for both systems, but does not list IPv6 on Linux in the current version. The other option would be to test for the available tool first. What would you prefer?

dequis:
Isn't /etc/default/ stuff ignored with systemd? Dropping the whole thing might be an option too. Or set it to 6667 and don't try to be clever about it.

jspricke:
Seems like, should I simply remove it?

jelmer:
Not trying to be clever about it seems reasonable to me too. It's nice to be consistent across init systems too.

jspricke:
I've updated the pull request accordingly.

jelmer:
You probably also want to update README.Debian to mention the new rather than the old behaviour.

PR #92, opened 2016-12-28
Use NI_MAXHOST rather than HOST_NAME_MAX for host lengths.

jelmer:
This constant is always available and meant to be used with
getnameinfo().

This fixes the build on Debian GNU/kFreeBSD. See https://buildd.debian.org/status/package.php?p=bitlbee


dequis:
Thanks. Almost forgot about this one.

PR #93, opened 2017-01-02
Debian stuff

dequis:
Last few systemd fixes, nuked /etc/default, nuked debconf, doc updates.

Stuff that i'm requesting review about: (@jelmer @wilm0r)

* that i'm not doing anything too crazy with the debconf thing
* the format and contents of the changelog (created with dch, but still not sure)
* these lintian complaints from `debuild -uc -us` in a debian sid box:

```
W: bitlbee source: native-package-with-dash-version
W: bitlbee source: diff-contains-git-control-dir .git
W: bitlbee source: debian-watch-file-in-native-package
```

Probably can be ignored (won't apply for proper packages)

```
W: bitlbee source: package-uses-deprecated-debhelper-compat-version 6
```

This warning is new, haven't seen it in older debians, and it looks like it complains about anything that isn't 9.

Right now it's 6 because of 560d0a0c4cc61e8ed68a80a04491af024254684c - regarding the dh_installdocs thing in that commit, the manpage says that adding that argument probably requires a migration so i'm not touching this.

```
W: bitlbee source: debian-rules-missing-recommended-target build-arch
W: bitlbee source: debian-rules-missing-recommended-target build-indep
W: bitlbee source: package-would-benefit-from-build-arch-targets
```

Probably just about adding makefile targets that point to build-stamp, sounds deceptively simple.

```
W: bitlbee-common: init.d-script-does-not-source-init-functions etc/init.d/bitlbee
```

Couldn't care less. Touching that one as little as possible.

```
W: bitlbee source: changelog-should-mention-nmu
W: bitlbee source: source-nmu-has-incorrect-version-number 3.5-1
```

These are a bit concerning, but maybe they don't apply once wilmer is the one doing the actual upload?


debian/bitlbee-common.config
```diff
@@ -1,21 +0,0 @@
-#!/bin/sh
-
-set -e
-
-. /usr/share/debconf/confmodule
-[ -f /etc/default/bitlbee ] && . /etc/default/bitlbee
-
-db_title BitlBee
-
-if [ -n "$BITLBEE_PORT" ]; then
```
jelmer: Won't this break existing users that rely on $BITLBEE_PORT? Ideally this would automatically be migrated, but at the very least, this kind of change needs a warning in NEWS.Debian.

jelmer: Same goes for settings this silently removes (e.g. BITLBEE_DISABLED="1")

debian/bitlbee-common.preinst
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
+set -e
+
+NAME=bitlbee
+PIDFILE=/var/run/$NAME.pid
+
+# git revision 7949d5a introduced dh_systemd.
+DH_SYSTEMD_REV="3.4.2+20160612+master+41-g7949d5a-git"
```
jelmer: dh-systemd is a developer tool, it won't be installed on users machines.

jelmer: There is no guarantee that the user will be using systemd. How would this work with e.g. sysvinit or upstart?

debian/bitlbee-common.config
dequis: Right. I was going to do that as suggested by #debian-devel, and half-forgot about it, half-thought that leaving a note in /etc/default/bitlbee itself is enough.

I actually like the idea of automatically migrating it, will give that a shot.

debian/bitlbee-common.preinst
dequis: Yeah, it's not used there, i just call this the "dh_systemd revision" because the packaging started using it, and the users probably have systemd units starting from that revision.

jelmer: What I mean is, dh_systemd being installed on a users' system doesn't say *anything* about whether they are running systemd. dh_systemd is a developer tool. It will be installed on some systems that are using sysvinit and won't be installed on most systems that are running systemd.

dequis: >There is no guarantee that the user will be using systemd. How would this work with e.g. sysvinit or upstart?

This particular preinst script doesn't depend on systemd, it uses the pidfile written by bitlbee to kill the master process using sysvinit-style commands. It's just to let systemd start the process normally because the sysvinit scripts didn't stop it.

dequis: ...these comment threads lack the auto-updating that is present everywhere else in github. welp.

jelmer: Argh, nevermind. The variable name confused me. You're comparing the bitlbee versions and not the dh_systemd versions. Ignore me.

debian/bitlbee-common.config
dequis: There, NEWS.Debian. That thing is decently loud, at least when upgrading.

I'm not sure about automatically migrating the port now. Editing the user config from scripts isn't going to be 100% reliable.

jelmer:
I think a mention in NEWS.Debian should be sufficient.

PR #95, opened 2017-01-20
Add --verbose configure option to control verbose build output

radhermit:
Defaults to disabled to maintain the status quo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,28 +162,28 @@ tar:
 	tar czf $$x.tar.gz --exclude=debian --exclude=.git* --exclude=.depend $$x
 
 $(subdirs):
-	@$(MAKE) -C $@ $(MAKECMDGOALS)
+	$(MAKE) -C $@ $(MAKECMDGOALS)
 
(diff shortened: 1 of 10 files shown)


dequis:
Yeah we needed something like this. I think it would be nicer to make it more like autotools:

```
  --enable-silent-rules   less verbose build output (undo: "make V=1")
  --disable-silent-rules  verbose build output (undo: "make V=0")
```

And set the real value of verbose (`@` or nothing) with an ifeq block based on `V`.

dequis:
Ooooon the other hand `--verbose=1` is more consistent with the rest of bitlbee's configure so uh maybe just do the `make V=1` thing? I like `make V=1`

radhermit:
How about something like that?

dequis:
Seems that I forgot to post my comment. Not a huge fan of having logic in Makefile.settings but I can't come up with anything better so it's fine. I'll merge this later.

jgeboski:
It's really not that much logic. This seems quite sane to me.

dequis:
Thanks! Sorry for the delay.

PR #97, opened 2017-02-13
Third argument to socket/socketpair fails on OS X

djnym:
I noticed that with bitlbee-3.5.1 that 'make check' would fail under Sierra.  After some investigation it appears that Sierra socket handling must have been made more strict.  To test, I ran this program under Linux and OS X.

```
#include <stdio.h>
#include <stdlib.h>

#include <sys/socket.h>

int main (int argc, char **argv) {
  int i;
  for (i = 0; i < 5; i++)
    {
      printf ("start with %d\n",i);
      if (socket (AF_UNIX, SOCK_STREAM, i) < 0) {
        perror("socket");
      }
    }
  exit(0);
}
```

Under Linux a protocol of ```0``` or ```1``` was successful, under OS X only ```0``` was successful.  According to the manpages, one should use ```0``` for the third argument (which is treated as using the default for the given family), and according to Stevens the only time a non-zero value makes sense is when using raw sockets.  I found that within bitlbee that sometimes ```0``` is used and other times (like in the IPC and test code) it was not.  This should make it consistent.

### DIFF
--- a/ipc.c
+++ b/ipc.c
@@ -897,7 +897,7 @@ int ipc_master_listen_socket()
 	un_addr.sun_family = AF_UNIX;
 	strcpy(un_addr.sun_path, IPCSOCKET);
 
-	serversock = socket(AF_UNIX, SOCK_STREAM, PF_UNIX);
+	serversock = socket(AF_UNIX, SOCK_STREAM, 0);
 
 	if (serversock == -1) {
 		log_message(LOGLVL_WARNING, "Unable to create UNIX socket: %s", strerror(errno));
--- a/tests/check.c
+++ b/tests/check.c
@@ -13,7 +13,7 @@ gboolean g_io_channel_pair(GIOChannel **ch1, GIOChannel **ch2)
 {
 	int sock[2];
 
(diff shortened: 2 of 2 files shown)


dequis:
>I couldn't find anyplace that actually enables the IPC, but I hand
edited the config.h to enable it, and did see the protocol not supported
failure, so modified this file and it went away.

`./configure --ipcsocket=...`

djnym:
That could be, but running ```./configure --help``` does not show that as an option, I had to actually look through the script to even see that it existed.  It looks like that script is hand constructed and that that option is not in the help string.  So maybe that's separate issue?

djnym:
Otherwise is this a reasonable patch, and is it possible to get a new version cut with it?  I want to get bitlbee via nixpkgs working and this was causing it to fail.  Thanks!

dequis:
Yeah I'll apply it later. A new version is unlikely. I suggest not running make check, it doesn't test useful stuff anyway.

dequis:
Thanks, sorry for the delay.

PR #98, opened 2017-03-15
irc: implement server-time capability

blastrock:
This closes https://bugs.bitlbee.org/ticket/1016

### DIFF
--- a/dcc.c
+++ b/dcc.c
@@ -213,7 +213,7 @@ int dccs_send_request(struct dcc_file_transfer *df, irc_user_t *iu, struct socka
 	cmd = g_strdup_printf("\001DCC SEND %s %s %u %zu\001",
 	                      df->ft->file_name, ipaddr, port, df->ft->file_size);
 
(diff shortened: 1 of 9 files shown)


dequis:
Neat, thanks!

Unfortunately, this breaks API (yeah, that stuff is API too - used by UI plugins like OTR. Yay for having everything exposed publicly). I've already been playing with implementing this (and the rest of message tags) in a safer way, just haven't had much time to finish it, but i'll get around to it soonish.

I'll keep this open meanwhile in case it's useful to anyone

blastrock:
Oh I didn't know that. Well I hope you get to finish this soon :)

dequis:
I figured it would be better to stop putting this off and merge it, since more people started merging it locally anyway. The ideal implementation can wait.

It's in the develop branch now, master later.

https://github.com/bitlbee/bitlbee/compare/f7cc734%5E...fa1bc

I just made some changes so that API isn't broken, and kept the new functions hidden so that when the more extensible approach is done later I can remove those.

PR #99, opened 2017-03-27
twitter: @mention the right people when replying

abrasive:
The previous approach was simply to @ the author of the replied-to
tweet. This is unintuitive
    a) when replying to a tweet that mentions
        several people, in which case they are cut off; and
    b) if you reply to a previous reply you made, in which case you @
        yourself and the intended recipient(s) won't see it.

This commit introduces a function twitter_format_reply to determine
the correct @mention string for the start of your tweet, based on the
original tweet you are replying to. This preserves all mentionees from
the original tweet, removing ourselves if present and making sure
the original tweet's author comes first (if not us).

This patch closes Trac issues #1289, #1251.

### DIFF
--- a/protocols/twitter/twitter.c
+++ b/protocols/twitter/twitter.c
@@ -670,6 +670,7 @@ static void twitter_login(account_t * acc)
 static void twitter_logout(struct im_connection *ic)
 {
 	struct twitter_data *td = ic->proto_data;
+	int i;
 
 	// Set the status to logged out.
 	ic->flags &= ~OPT_LOGGED_IN;
@@ -699,6 +700,11 @@ static void twitter_logout(struct im_connection *ic)
 		g_free(td->prefix);
 		g_free(td->url_host);
 		g_free(td->url_path);
+		for (i=0; i<TWITTER_LOG_LENGTH; i++) {
+			if (td->log[i].text) {
+				g_free(td->log[i].text);
+			}
+		}
 		g_free(td->log);
 		g_free(td);
 	}
@@ -922,6 +928,67 @@ static guint64 twitter_message_id_from_command_arg(struct im_connection *ic, cha
 	return id;
 }
 
+/** Find the log entry holding a given tweet ID, or NULL if absent or the user object is dangling.
+ */
+static struct twitter_log_data * twitter_find_log(struct twitter_data *td, guint64 id)
+{
+	int i;
+	for (i=0; i<TWITTER_LOG_LENGTH; i++) {
+		if (td->log[i].id == id) {
+			return &td->log[i];
+		}
+	}
+
+	return NULL;
+}
+
+static gchar * twitter_format_reply(struct im_connection *ic,
+                                    guint64 id, bee_user_t *bu, char *message)
+{
+	struct twitter_data *td = ic->proto_data;
+	struct twitter_log_data *ld = twitter_find_log(td, id);
+
+	if (!ld) {
+		return g_strdup_printf("@%s %s", bu->handle, message);
+	}
+
+	GString *reply = g_string_new(NULL);
+	// Always mention the replied-to tweet's author first - if not us
+	if (bu != &twitter_log_local_user) {
+		g_string_printf(reply, "@%s ", bu->handle);
+	}
+
+	gchar *orig = g_strdup(ld->text);
+	gchar *op = g_strchug(orig);
+	gchar *next;
+	while (op && *op == '@') {
+		char *sep = strchr(op, ' ');
+		if (sep) {
+			*sep = '\0';
+			next = sep + 1;
+		} else {
+			next = NULL;
+		}
+
+		// Add mention to reply, if not us or already done
+		if (g_ascii_strcasecmp(op + 1, twitter_log_local_user.handle) &&
+			g_ascii_strcasecmp(op + 1, bu->handle)) {
+			g_string_append(reply, op);
+			g_string_append_c(reply, ' ');
+		}
+
+		op = next;
+		while (g_ascii_isspace(op)) {
+			op++;
+		}
+	}
+
+	g_free(orig);
+
+	g_string_append(reply, message);
+	return g_string_free(reply, FALSE);
+}
+
 static void twitter_handle_command(struct im_connection *ic, char *message)
 {
 	struct twitter_data *td = ic->proto_data;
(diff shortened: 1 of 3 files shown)


dgw:
This is the number-one reason I find myself breaking out of bitlbee's Twitter implementation (needing to reply to my own tweet or replying to a tweet that mentions multiple people)—I can't give a big enough 👍

dequis:
>needing to reply to my own tweet

There's a rawreply command btw

Wilm0r:
What dx says .The reason reply insists to put the mention at the start is because, annoyingly, Twitter used to require it (they'd discard the in_reply_to field if there was no matching mention at the start) .... just later changed their mind on that. :<

It's annoying because changing the behaviour of the reply command is a somewhat annoying bw compatibility change. :-(

I'm not super enthusiastic about caching full tweets just to make the additional mentions work. I don't always want them, even in the web UI I often remove a few to get back some precious character quota...

ticky:
The trouble is that the _current_ behaviour means that a user can unknowingly start a chain of replies which inadvertently don’t actually mention the person the user thinks they’re replying to.

“rawreply” isn’t really the best user-facing solution to that when you have a computer which can do the work for you!

dgw:
> “rawreply” isn’t really the best user-facing solution to that when you have a computer which can do the work for you!

This. Being able to accomplish the goal by doing the work myself when the computer can do the work for me (and faster) is sub-optimal. `rawreply` would still be useful for omitting mentions to regain characters if desired.

> The reason reply insists to put the mention at the start is because, annoyingly, Twitter used to require it (they'd discard the in_reply_to field if there was no matching mention at the start) .... just later changed their mind on that. :<

So even if Twitter changes their mind, BitlBee must stick to the old way forever? It might not be appropriate for a patch release, but at the next "big enough" version bump, this behavior change could be included and called out in the release notes.

Obviously I'm biased toward including it at some point, rather than never doing this and always mentioning the replied-to tweet's author no matter what, even if it's myself… but if the behavior change can be made at an appropriate time, then the only remaining objections (as I see it) are to the specific implementation. That can be changed to, for example, cache a smaller amount of data instead of full tweets.

ticky:
It may also be of note that the Twitter API also supplies the users mentioned in a tweet in [`entities.user_mentions`](https://dev.twitter.com/overview/api/entities), which could be cached instead of the tweet text if optimisation is a priority.

dequis:
I only suggested rawreply for a single use case, replying to yourself. It works for others but it's not user friendly. I think I'm fine with reply having unpredictable (but hopefully expected) behavior and rawreply always doing exactly what the user asked.

>which could be cached instead of the tweet text if optimisation is a priority

Probably takes more space to store those structures than to store the raw text.

Not overly concerned about memory usage myself

protocols/twitter/twitter.c
```diff
@@ -699,6 +700,9 @@ static void twitter_logout(struct im_connection *ic)
 		g_free(td->prefix);
 		g_free(td->url_host);
 		g_free(td->url_path);
+		for (i=0; i<TWITTER_LOG_LENGTH; i++)
+			if (td->log[i].text)
```
dequis: Please use braces in all blocks, even when they can be omitted. This applies to other places too.

abrasive:
Fixed the brace style in the patch, sorry about that.

dequis:
I should mention, the `auto_populate_reply_metadata` parameter exists and does something very similar to this, with the advantage that we don't store anything and mentions aren't counted towards the tweet body length limit. It's documented here https://dev.twitter.com/overview/api/upcoming-changes-to-tweets

However, I'm uncertain on what will happen to this feature - they took the unusual approach of providing it first through the API and taking care of UI later, but no one seems to like the UI side of their implementation (which involves hiding leading mentions) and they've been turning it on and off for different a/b testing groups for a while. It might make sense to assume that since it's part of the documented API they are committed to support it... but I don't know.

abrasive:
Is it practical to detect a failure with `auto_populate_reply_metadata` and fall back to this method? The improved length limit would be lovely to have, but I'm as leery of Twitter's API behaviour as you are; having to rush out a rollback when they break it would be a bit unpleasant.

abrasive:
Welp, looks like this patch is now obsolete. Does this mean `auto_populate_reply_metadata` is now required?

dequis:
No, It's still optional and keeping it disabled is desirable in some cases (for example, i wouldn't enable it for rawreply). The rules on the API side haven't changed, it's still the same as that document, the only change is that they deployed the twitter.com UI.

I'm not quite sure what's the right way to go with this whole thing, but this patch is still useful for statusnet/gnu social accounts, which are also supported by bitlbee since their api is twitter-compatible (but frozen in time).

And now it's more important than before to be able to detect how far the twitter compatibility of a server goes. I'm thinking maybe it's time to use `/1.1/help/configuration.json`

ticky:
The Twitter dot com update changes nothing about how you compose a tweet in a text-only environment (read: bitlbee, other third-party twitter clients). This patch is still legit!

dequis:
@ticky it's not like this patch allows customizing the set of people who are mentioned either. The behavior of this patch is (as far as i know) the same as what's offered by `auto_populate_reply_metadata` flag, which is what the twitter.com UI started using today.

The part that looks like it requires a complex UI is the `exclude_reply_user_ids` parameter. But we could just not implement that.

Wilm0r:
Yeah, this is rather unexpected timing of things @ Twitter. But maybe about time because just piling up the mentions in comment threads and that actually eating into your character budget also felt much like taking the piss.

And oh yes, mostly I was providing background on why reply does this thing that seems stupid but was simply required at the time. Also on memory consumption: Yeah it's not like BitlBee runs on embedded systems with their RAM measured in KB, but all parts of it were written as much as possible as a dumb relay that keeps very little state, and this change would fall out a little bit.

And as dx points out, just including all mentions with little automatic control isn't optimal either - the problem here is that a proper client has its own line editor which is pre-populated with all the suggested mentions and you can modify them to taste. Without that, I don't know whether automatically (and with the default command for that purpose) spending half of the character budget (if that were still a thing) would be reasonable..

flexo:
Would it instead be possible to have something like a greply (group reply) command that did add all the recipients? Email makes the differentiation between Reply Sender and Reply All so it'd at least be a familiar concept.
Could even have a setting that allowed users to choose between reply and greply as the default behaviour in the 'nick: reply text' syntax.

Wilm0r:
This has always been a little "difficult" because of the way BitlBee
keeps state (preferably as little as possible), allowing for at most one
handle and only one that is in your contact list.

But I wonder whether with the new API with the mentions kinda oob, we
could just specify a flag and have Twitter automatically include them
all.

rixx:
A feature like what @iguanaonmystack suggested, `greply`, sounds pretty good to me and wouldn't be a compatibility change. For me, the bitlbee Twitter reply behaviour is making Twitter very hard to use, because my primary usecase are conversations with multiple friends.

Is there a chance/hope/way of moving this patch forward?

dequis:
This patch is in a really awkward place given `auto_populate_reply_metadata`...

dequis:
Merged #146 instead, closing this, sorry!

PR #100, opened 2017-04-03
Add PROXY command so stunnel connections look better

Wilm0r:
Tired of all the "localhost" connections on public servers. There are horrible root-requiring hacks that make stunnel a transparent proxy, I'd rather just do this.

There's a theoretical chance of someone using this to spoof, but they'd need local-machine access so meh. It's not like this controls any ACLs..

Also, add verification of reverse lookups, so far all PTR lookup results were just used blindly.

### DIFF
--- a/lib/canohost.c
+++ b/lib/canohost.c
@@ -1,0 +1,140 @@
+  /********************************************************************\
+  * BitlBee -- An IRC to other IM-networks gateway                     *
+  *                                                                    *
+  * Copyright 2002-2017 Wilmer van der Gaast and others                *
+  \********************************************************************/
+
+/*
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
+  the Free Software Foundation; either version 2 of the License, or
+  (at your option) any later version.
+
+  This program is distributed in the hope that it will be useful,
+  but WITHOUT ANY WARRANTY; without even the implied warranty of
+  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  GNU General Public License for more details.
+
+  You should have received a copy of the GNU General Public License with
+  the Debian GNU/Linux distribution in /usr/share/common-licenses/GPL;
+  if not, write to the Free Software Foundation, Inc., 51 Franklin St.,
+  Fifth Floor, Boston, MA  02110-1301  USA
+*/
+
+/* These two functions, with some modifications, taken from OpenSSH:
+ *
+ * $OpenBSD: canohost.c,v 1.73 2016/03/07 19:02:43 djm Exp
+ *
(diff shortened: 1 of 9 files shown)


dequis:
reverse_lookup() really should be in a separate file with the original copyright string and license text from the file it was taken from. `debian/copyright` is not available in all places.

This should probably be disabled by default in bitlbee.conf.

There's a WEBIRC command which has been around for years, and it recently got a clean version of the spec proposed for inclusion in ircv3 (https://github.com/ircv3/ircv3-specifications/pull/303 - should have been merged by now imo). Anyway it's a lot like PROXY with an optional password, but I'm not sure how useful it is if stunnel doesn't support it.

Also the haproxy doc link is down. So much for high availability. Ha. ok sorry

dequis:
>should have been merged by now imo

And it actually got merged as I was writing that comment. That's neat.

Wilm0r:
canohost.[ch] added.

Yeah, WEBIRC looks nicer, but indeed stunnel doesn't support it and I don't think that could easily be added. (Easy as-in, say, without recompiling, by for example just giving it a format string or something.) Makes it less attractive to me. :-(

Care much about the setting? I've thought of it, but since no ACLs or whatever depend on this and since I'm probably the only person on earth who sometimes uses OPER for something other than "secure" password entry, I can't really get my caring level up enough to make it configurable.

lib/canohost.c
```diff
@@ -0,0 +1,140 @@
+  /********************************************************************\
+  * BitlBee -- An IRC to other IM-networks gateway                     *
+  *                                                                    *
+  * Copyright 2002-2017 Wilmer van der Gaast and others                *
+  \********************************************************************/
+
+/*
+  This program is free software; you can redistribute it and/or modify
+  it under the terms of the GNU General Public License as published by
```
dequis: lol

dequis:
>Yeah, WEBIRC looks nicer, but indeed stunnel doesn't support it and I don't think that could easily be added. (Easy as-in, say, without recompiling, by for example just giving it a format string or something.) Makes it less attractive to me. :-(

Yeah forget about webirc. I briefly considered adding it to stunnel myself but it doesn't use any version control system. None at all. They just put up a release tarball every time they want to change something. Literally worse than bzr.

Also I heard from others that they implemented this thing in their ircds at least internally

>Care much about the setting? I've thought of it, but since no ACLs or whatever depend on this and since I'm probably the only person on earth who sometimes uses OPER for something other than "secure" password entry, I can't really get my caring level up enough to make it configurable.

Okay. I think the checks to only accept this from localhost are enough.

lib/canohost.c
Wilm0r: This *program*, not this file. It's the default file header, I flipped a coin and it told me to keep it. ;-)

PR #102, opened 2017-05-19
Embed images support in libpurple

Roger:
Parse  tags in messages, store the image in the filesystem and
replace the tag with a url to the image.

This add two new settings, webserver_root and webserver_url, and only
stores images if they exist.

The user should configure an http server, pointing to the webserver_root
and serving to webserver_url

Example Config:
webserver_url = http://example.com
webserver_root = /srv/bitlbee/

### DIFF
--- a/protocols/purple/purple.c
+++ b/protocols/purple/purple.c
@@ -1212,6 +1212,7 @@ replace_images_cb (const GMatchInfo *info, GString *res, gpointer data)
 
 	g_free(id);
 	g_free(path);
+	purple_imgstore_unref(image);
 	return FALSE;
 
 error:
--- a/protocols/purple/purple.c
+++ b/protocols/purple/purple.c
@@ -1174,13 +1174,66 @@ void prplcb_conv_del_users(PurpleConversation *conv, GList *cbuddies)
 	}
 }
 
+/* Store images and replace <img> tags with urls pointing to them */
+static gboolean
+replace_images_cb (const GMatchInfo *info, GString *res, gpointer data)
+{
+	gchar *id;
+	const gchar *ext;
+	guchar *imgdata;
+	size_t imglen;
+	gchar *hash;
+	gchar *path;
+
+	if (global.conf->web_directory == NULL || global.conf->web_url == NULL) {
+		goto error;
+	}
+
+	id = g_match_info_fetch(info, 1);
+	PurpleStoredImage *image = purple_imgstore_find_by_id(atoi(id));
+
+	if (!image) {
+		goto error;
+	}
+
+	ext = purple_imgstore_get_extension(image);
+	imgdata = (guchar *) purple_imgstore_get_data(image);
+	imglen = purple_imgstore_get_size(image);
+
+	hash = g_compute_checksum_for_data(G_CHECKSUM_SHA256, imgdata, imglen);
+	path = g_strdup_printf("%s/%s.%s", global.conf->web_directory, hash, ext);
+
+	// only save images that don't exist in the filesystem, in protocols like telegram lots of images are duplicated(stickers)
+	if (access(path, F_OK) == -1) {
+		g_file_set_contents(path, (gchar *) imgdata, imglen, NULL);
+	}
+
+	g_string_append_printf(res, "%s/%s.%s", global.conf->web_url, hash, ext);
+
+	g_free(id);
+	g_free(path);
+	return FALSE;
+
+error:
+	g_string_append(res, "[image]");
+	return FALSE;
+}
+
+
 /* Generic handler for IM or chat messages, covers write_chat, write_im and write_conv */
 static void handle_conv_msg(PurpleConversation *conv, const char *who, const char *message_, guint32 bee_flags, time_t mtime)
 {
 	struct im_connection *ic = purple_ic_by_pa(conv->account);
 	struct groupchat *gc = conv->ui_data;
-	char *message = g_strdup(message_);
 	PurpleBuddy *buddy;
+	static GRegex *regex = NULL;
+	gchar *message;
+
+	if (regex == NULL) {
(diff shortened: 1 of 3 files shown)


dequis:
I reviewed this thing out-of-band with roger before he opened the PR so the code already looks good to me and I'd merge it. Probably adding some documentation afterwards, like including the new settings in bitlbee.conf

But I would like to hear opinions from the others (@Wilm0r @sm00th @jgeboski @EionRobb) regarding the "web_directory" and "web_url" global settings

(btw @Roger your commit log got the names wrong, fix that)

The idea has been suggested before by seirl in [trac ticket 1192](https://bugs.bitlbee.org/ticket/1192) ("Leave the http server responsibility to the user, and display an URL formatted like the user configured it")

And spectrum2 (which is like bitlbee for xmpp) did this before too, see https://github.com/hanzz/spectrum2/commit/65d5b4fe996f1a137687e55dc0d0298ac4d21568

It's disabled by default so that's safe. There's no size limit (spectrum has 1mb limit, which sounds too low - probably worth adding a third setting to do that). There's also no maximum size limit so running out of disk space is a real risk, which we could document and maybe later provide some ways to help keep it limited.

There's no structure under web_directory, just filenames that are "sha256.ext". That's probably ok, not sure if i'd want to add even more settings, and being in a single place helps with deduplication (files sent more than once are not written more than once)

Anyway, just looking for opinions on the general concept before introducing this as a new feature.

protocols/purple/purple.c
```diff
@@ -1174,13 +1174,67 @@ void prplcb_conv_del_users(PurpleConversation *conv, GList *cbuddies)
...
-	char *message = g_strdup(message_);
 	PurpleBuddy *buddy;
+	static GRegex *regex = NULL;
+	gchar *message;
+
+	if (regex == NULL) {
+		regex = g_regex_new("<img id=\"(\\d+)\">", 0, 0, NULL);
+	}
+
+	message = g_regex_replace_eval(regex, message_, -1, 0, 0, replace_images_cb, NULL, NULL);
```
EionRobb: Instead of using a regex, use `purple_markup_find_tag()` which processes xml and handles the case of that string you're trying to match being embedded in other strings

protocols/purple/purple.c
```diff
@@ -1174,13 +1174,67 @@ void prplcb_conv_del_users(PurpleConversation *conv, GList *cbuddies)
...
+
+	ext = purple_imgstore_get_extension(image);
+	imgdata = (guchar *) purple_imgstore_get_data(image);
+	imglen = purple_imgstore_get_size(image);
+
+	hash = g_compute_checksum_for_data(G_CHECKSUM_SHA256, imgdata, imglen);
+	path = g_strdup_printf("%s/%s.%s", global.conf->web_directory, hash, ext);
+
+	// only save images that don't exist in the filesystem, in protocols like telegram lots of images are duplicated(stickers)
+	if (access(path, F_OK) == -1) {
```
EionRobb: The man page says to not ever use `access()` in production.  Perhaps use `g_file_test()` instead

protocols/purple/purple.c
```diff
@@ -1174,13 +1174,67 @@ void prplcb_conv_del_users(PurpleConversation *conv, GList *cbuddies)
...
+		g_file_set_contents(path, (gchar *) imgdata, imglen, NULL);
+	}
+
+	g_string_append_printf(res, "%s/%s.%s", global.conf->web_url, hash, ext);
+
+	g_free(id);
+	g_free(path);
```
EionRobb: You've got a memleak on `hash`, you need to `g_free()` it

protocols/purple/purple.c
dequis: >The man page says to not ever use `access()` in production

What version of the man page and where? `g_file_test()` uses `access()` (for the unix part, and we don't care about the win32 part here. sorry eion). Probably a good idea to use `g_file_test()` anyway.

EionRobb: https://linux.die.net/man/2/access in the notes "For this reason, the use of this system call should be avoided"

protocols/purple/purple.c
dequis: Maybe, but that's just the finding part. The replacement part requires doing fiddly pointer arithmetic. I'm not sure it's worth it if this is just for the edge case where someone does `<tag attr=''>`

protocols/purple/purple.c
dequis: Oh, the time-of-check-time-of-use vuln. That's not about the system call itself, it's about how it's used (`g_file_test()` can be used incorrectly like that too, [the docs mention it][1]). We're not checking for permissions here, just for existence, so the risk is that a file with the same name is created between the check and our overwrite. Non-issue imo.

[1]: https://developer.gnome.org/glib/stable/glib-File-Utilities.html#g-file-test

rodneyrod:
What's the progress on this one?

Roger:
I have some code done.. but had no time to finish it

Roger:
This is the last I was working, if someone want to try to finish it

https://github.com/Roger/bitlbee/commit/a04c222adc8722f9102e38af230507ad31b8436e

Thaodan:
What about remote access? Do you store image data per ident?

Roger:
> What about remote access? Do you store image data per ident?

The remote access should be configured as an external http server serving the images directory, the images are stored in one directory using the hash and the extension only, not per ident

Thaodan:
So the hash can't conflict with user other on 
the bitlbee instance?
Should the external webserver include the 
auth data like using the ident name as a 
user name for http auth?

protocols/purple/purple.c
Thaodan: Is `1purple_imgstore_unref(image)` still needed?

PR #103, opened 2017-06-29
Fix build without python installed

rbgarga:
Since help.txt is part of the tarball and python is not needed it
doesn't need to stop configure script when --doc=1 and python is not
present in the system

### DIFF
--- a/configure
+++ b/configure
@@ -763,20 +763,8 @@ if [ "$skype" = "1" -o "$skype" = "plugin" ]; then
 	protocols_mods="$protocol_mods skype(plugin)"
 fi
 
-if [ -z "$PYTHON" ]; then
-	PYTHON=python
-fi
-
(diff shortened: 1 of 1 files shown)


dequis:
>Since help.txt is part of the tarball

It is, but it's not in git versions and it has to be generated at some point. If you remove PYTHON from Makefile.settings it can never be generated.

rbgarga:
It make sense. May we should have a different target used decide if you want or not to rebuild help.txt, and then, if it's selected, add python dependency. My first attempt was to use --doc=0 but then I ended up without manpages.

dequis:
Solved with 4d51c8451c6abb2fa90e099034fcc9c39e7530aa, thanks for reminding me about this issue.

rbgarga:
Thank you!

PR #105, opened 2017-08-09
Twitter: fix format strings for integers

kensanata:
Silence the following compiler warning: format specifies type
'unsigned long long' but the argument has type 'guint64'. When
formatting a json_integer, use PRId64 for int64_t. When formatting
ordinary integers, use G_GUINT64_FORMAT for guint64.

### DIFF
--- a/protocols/twitter/twitter_lib.c
+++ b/protocols/twitter/twitter_lib.c
@@ -307,7 +307,7 @@ static gboolean twitter_xt_get_friends_id_list(json_value *node, struct twitter_
 		}
 
 		txl->list = g_slist_prepend(txl->list,
-		                            g_strdup_printf("%" PRIu64, c->u.array.values[i]->u.integer));
+		                            g_strdup_printf("%" PRId64, c->u.array.values[i]->u.integer));
 	}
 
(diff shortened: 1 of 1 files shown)


dequis:
`G_GINT64_FORMAT`?

kensanata:
If we use guint64 we should use [G_GUINT64_FORMAT](https://developer.gnome.org/glib/stable/glib-Basic-Types.html#G-GUINT64-FORMAT:CAPS), I think.

dequis:
I mean `G_GINT64_FORMAT` for `int64_t`. There's no reason to use the portable glib macro for the unsigned type and not for the signed one.

kensanata:
Tried it, and I get "warning: format specifies type 'long' but the argument has type 'int64_t' (aka 'long long')". I'm not well versed understand what's going on, all I have is an aesthetic reason: if we use a type from a certain family, we should use the format string specified for the same family. And being a Stack Overflow kind of person, I just looked at what [other people are saying](https://stackoverflow.com/questions/9225567/how-to-print-a-int64-t-type-in-c) which links to some [C++ reference](http://en.cppreference.com/w/cpp/types/integer). That's all I know.

PR #108, opened 2017-08-31
Add datadir to pkconfig file

sm00th:
Export datadir through pkgconfig for later use by plugins.


dequis:
Typo in the commit log "pkconfig"

Maybe worth including as a constant in config.h too?

sm00th:
> Typo in the commit log "pkconfig"

fixed, thanks

> Maybe worth including as a constant in config.h too?

Oh neat-o. Didn't know it was there (although it suffers the same problem as json.h, its a pain to include with `-I/usr/include/bitlbee`)

PR #110, opened 2017-10-12
Handle always_use_nicks more gracefully

Chemrat:
- Add MUC buddies before applying nick changes
- Rename buddies on nick change instead of forcing part/join

This should apply nicks to all MUC clients, not just people on your contact list
Also renames just look better than part/join

I'm not too familiar with bitlbee's inner workings, but I've been running bitlbee with this change for a couple of days now and it seems to work for me

### DIFF
--- a/protocols/jabber/conference.c
+++ b/protocols/jabber/conference.c
@@ -350,11 +350,12 @@ void jabber_chat_pkt_presence(struct im_connection *ic, struct jabber_buddy *bud
 			*s = 0; /* Should NEVER be NULL, but who knows... */
 		}
 
+		imcb_chat_add_buddy(chat, bud->ext_jid);
+
 		if (bud != jc->me && (jc->flags & JCFLAG_ALWAYS_USE_NICKS) && !(bud->flags & JBFLAG_IS_ANONYMOUS)) {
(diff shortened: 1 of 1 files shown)


dequis:
Neat. Will look at this later. I think `imcb_buddy_nick_change` didn't exist when I wrote that MUC nick changing thing.

dequis:
Hi, sorry for the delay.

I applied the first half of this (Add MUC buddies before applying nick changes) as f75aaac5e9c0a0ce6e19c3d167ae41acc02f95b1

It works as intended and I can confirm it fixes one `always_use_nicks` bug, hopefully the only one.

The second half (Rename buddies on nick change instead of forcing part/join) doesn't work - it tries to do the rename when it gets the unavailable presence, which doesn't include the new nick, so it attempts to rename from old nick to old nick. The new nick is included in a separate presence stanza. The effect of this change right now is that the quit disappears and a new user is joined to the channel.

So getting that one right will take some more effort (possibly a flag in the jabber buddy telling it that it's being renamed, to be handled later when the available presence is received?) and I don't think it's worth the hassle, but if you want to spend some time on it feel free to open another PR.

Thank you!

PR #111, opened 2018-02-06
Update Twitter character limit to 280 characters

alyssais:


rodneyrod:
Quick question, is 280 now the default standard for all accounts in all languages or are there accounts with different maximum character limits?

dgw:
It's a good question. I was poking around Twitter's documentation before updating the character counter in YOURLS's share function but they really don't want to talk about character limits any more. They just say the limit is "[more flexible](https://developer.twitter.com/en/docs/basics/counting-characters)" now.

280 is a good rule of thumb, I suppose. It's nominally what users can expect, at least when tweeting in alphabetic languages.

dequis:
Yeah so one reason I've been putting this off is that 280 isn't accurate, it's 280 with weighting. The weighting is defined in a way that (almost) all characters after U+10FF (except a handful of code points in the 'General Punctuation' block) have double weight. So it's not a per-account thing but if you write japanese the counting will effectively get you 140.

This is somewhat documented [here][1], although the canonical documentation is probably the twitter-text library by now.

twitter-text describes this stuff with JSON config, stored client side, with no obvious way to fetch from the server. My idea was to implement equivalent config parsing and make it as easy as possible to just copy-paste twitter-text-formatted json configs into bitlbee.

But realistically I have too much shit to do and that one isn't a priority.

So yeah I guess we'll just do this.

[1]: https://developer.twitter.com/en/docs/developer-utilities/twitter-text.html

dequis:
One thing though: can you rebase to develop so I don't drop the fancy verified thing? It's one commit ahead.

alyssais:
Rebased!

dequis:
Thanks!

PR #114, opened 2018-07-03
Stop using the irc->users linked list, use the hash table instead

dequis:
~~[INCOMPLETE! Some things are half-assed. But it works.]~~

irc_user_new() mentions that the reason this list is kept is for easy
iteration. Luckily, this is the future, and GHashTableIter exists now.

The main point of this is to get rid of the g_slist_insert_sorted() in
irc_user_set_nick() which is a particularly slow part of loading large
user lists, and scales poorly

In a test with discord, the GUILD_SYNC event is now 4 times faster, on
top of the improvements of the other bee_user hash tables patch.
Combining both patches it went from 136 to 6 seconds for 50k members.

-----

Callgraph that led to this:

sari4

Callgraph after:

vbkh4

That stuff is handled at https://github.com/sm00th/bitlbee-discord/pull/165

### DIFF
--- a/irc.c
+++ b/irc.c
@@ -241,6 +241,8 @@ static gboolean irc_free_hashkey(gpointer key, gpointer value, gpointer data);
 void irc_free(irc_t * irc)
 {
 	GSList *l;
+	GHashTableIter iter;
+	gpointer itervalue;
 
 	irc->status |= USTATUS_SHUTDOWN;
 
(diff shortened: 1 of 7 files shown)


dequis:
TODO:

- [ ] `irc_user_free` modifies the hash table and so invalidates the iter, resulting in a warning at exit. Didn't seem to crash but the docs say it's not safe.
- [ ] `blist` iteration is now unordered.
- [ ] Check whether those `g_hash_table_contains` make sense (key vs value)
- [ ] That commented out line in `irc_send_who`
- [ ] Test

dequis:
Addressed my TODOs, done I think. Regarding sorting `blist` output:

>04:52 < dx> i just re-added sorted output to the blist command. it's slower (200ms vs 90ms with 14k users), but only if you comment out the irc_rootmsg() calls, because those take several seconds to build the send buffer, making the other slowdown unnoticeable

I sneakily threw b9d1fdc7208e7329122e76e1aa67384442c1cfae to develop to improve that slightly, but the real improvement of send buffers is in #117 

I'll merge both of these later

PR #115, opened 2018-07-09
show correct nick when `rename -del` is used

arcnmx:
I thought `rename -del` was broken until I looked at the source and realised it was just trying to tell me that the nick I was deleting didn't exist. Fixes this error message:

    <root> Nick `-del' does not exist

Tangential, but `help rename` suggesting `rename -del <oldnick>` is a bit misleading and implies that the original handle should be used with the command, instead of the current/new/renamed nick. Would just `<nick>` be more appropriate there maybe?

EDIT: aaand I just realised this is targeting the wrong branch. can I change that ;-;

### DIFF
--- a/root_commands.c
+++ b/root_commands.c
@@ -818,7 +818,7 @@ static void cmd_rename(irc_t *irc, char **cmd)
 	iu = irc_user_by_name(irc, cmd[del ? 2 : 1]);
 
 	if (iu == NULL) {
-		irc_rootmsg(irc, "Nick `%s' does not exist", cmd[1]);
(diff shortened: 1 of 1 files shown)


dequis:
```
14:33 < Not-a9c> [bitlbee] arcnmx opened pull request #115: show correct nick when `rename -del` is used - https://git.io/fNJBW
14:34 < krisfremen> That's new
14:39 < Not-a9c> [bitlbee] arcnmx edited pull request #115: show correct nick when `rename -del` is used - https://git.io/fNJBW
14:43 < dx> that's mew
```

arcnmx:
mew~!

<sub>where's my mew reaction github ;;</sub>

dequis:
Thanks~!

PR #116, opened 2018-07-09
rename -del help clarification

arcnmx:
In context next to `help rename <oldnick> <newnick>`, instructing users to `rename -del <oldnick>` indicates that the original handle should be used instead of the user's current nick. This is not the case, so I think this change makes sense to help clarify that?

This could also be `rename -del <newnick>` instead to indicate that it's for deleting nicks that have already been renamed using the aforementioned command? Either way works for me.


dequis:
Thanks~!

PR #118, opened 2018-09-09
SSL: correctly handle hostnames starting with a digit when SNI is needed.

CMB:
I noticed that I was unable to connect to a host with bitlbee.
Tracking it down, I found that the problem was that the host needed SNI
but bitlbee wasn't trying to use it.  The hostname started with a digit,
so bitlbee was assuming that it was an IP address.


dequis:
Thanks!

PR #124, opened 2019-03-31
Add AVATAR to XMPP

SoniEx2:
http://www.kvirc.net/doc/doc_ctcp_avatar.html

The CTCP AVATAR is "AVATAR URI-or-filename optional-filesize-hint"

It supports HTTP, HTTPS and DCC, with HTTP and HTTPS preferred.

Additionally, for users with AVATAR, one needs to insert a hidden color code 4 immediately followed by a formatting reset at the start of the realname (i.e. "\x034\x0f"). This tells KVIrc that the user has an avatar.

(btw: trac is broken. every time I try to submit an issue/create an account/etc it just crashes and times out. if that can't be fixed, can you at least open the github issue tracker?)


dequis:
Don't do this.

PR #127, opened 2019-04-21
debian: fix up copyright file

dequis:
For https://bugs.debian.org/883872

----

Several entries are just gone because the oscar plugin is gone.


dequis:
cc @mtecknology

PR #128, opened 2019-05-23
conf: support AccountStorage as well as Account_Storage

avar:
These variables added way back in b73ac9c3 ("Add support for 'primary'
and 'migrate' account storages. [...]", 2005-12-14). I think that
these have never been used by anyone, I think Dennis's out-of-tree
MySQL backend is the only storage backend anyone bothered to write.

That backend uses "AccountStorage" in the config, not
"Account_Storage". Let's support that, we could just change this over,
but *maybe* someone uses it out of tree, and supporting their config
in that case is easy.

I'm forward-porting https://github.com/bitlbee/bitlbee/pull/67 there's
production code that uses the CamelCase variant, and it would be handy
if it worked as-is without needing to fiddle with the config.

### DIFF
--- a/conf.c
+++ b/conf.c
@@ -276,10 +276,12 @@ static int conf_loadini(conf_t *conf, char *file)
 			} else if (g_strcasecmp(ini->key, "motdfile") == 0) {
 				g_free(conf->motdfile);
 				conf->motdfile = g_strdup(ini->value);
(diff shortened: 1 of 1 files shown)


avar:
@seveas : FYI.

dequis:
Thanks!

PR #129, opened 2019-05-23
Don't make early warnings go nowhere

avar:
between log_init and the existing log_link calls, all errors and
warnings would go nowhere. Let's show them on the console.


avar:
@seveas : FYI.

dequis:
Thanks!

PR #130, opened 2019-05-23
Centralize password handling functions

avar:
In preparation for the next commit and the forthcoming MySQL storage
backend, I needed to centralize all password handling. This commit
achieves a few things at once that are not easy to split up in multiple
commits while keeping every commit buildable/testable:

- One place to encode/encrypt/decode/decrypt/verify all passwords
- Less code duplication (several places did their own md5 handling)
- Reusable password encryption
- Generic function names instead of hardcoding md5/arc in the name
- Clearer and easier to extend password checking logic

### DIFF
--- a/bitlbee.h
+++ b/bitlbee.h
@@ -146,6 +146,7 @@ extern "C" {
 #include "sock.h"
 #include "misc.h"
 #include "proxy.h"
+#include "password.h"
 
 typedef struct global {
 	/* In forked mode, child processes store the fd of the IPC socket here. */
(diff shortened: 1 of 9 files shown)


avar:
This is the first bit of @seveas's 2016-era changes previously submitted in https://github.com/bitlbee/bitlbee/pull/67. Along with #128 and #129 I'm submitting those piecemeal for ease of review.

avar:
For things I'm trying to upstream which aren't https://github.com/bitlbee/bitlbee/commit/5b9bbb14 I see this isn't needed. I.e. it's prep work for that, but I'm not interested in changing out the password format. So just closing this, sorry.

PR #133, opened 2019-08-25
Fix lintian issues: Add missing colon in closes line., Bump debhelper from old 10 to 12., Remove obsolete fields Name from debian/upstream/metadata., Use secure URI in Homepage field., Use secure URI in debian/watch.

jelmer:
Fix some issues reported by lintian
* Use secure URI in debian/watch.
* Use secure URI in Homepage field.
* Bump debhelper from old 10 to 12.
* Add missing colon in closes line.
* Remove obsolete fields Name from debian/upstream/metadata.


dequis:
Travis is complaining and I don't quite get it, but it says they are killing the ubuntu 12.04 environment soon, which sounds fair. Would updating that help?

jelmer:
Yeah, that would help - this appears to be failing due to an old version of debhelper that doesn't support the debhelper-compat syntax (https://nthykier.wordpress.com/2019/01/04/debhelper-compat-12-is-now-released/)

Note that the current stable release of Debian (buster) supports debhelper 12.

dequis:
Hm, but I'd like the build to succeed in all currently supported versions of ubuntu and debian. Is ubuntu 16.04 enough? Otherwise we might want to keep the older debhelper compat.

jelmer:
Nope, xenial has debhelper 9 (https://packages.ubuntu.com/search?keywords=debhelper). Do you allow backports? Because bitlbee is already using debhelper 10, and that would mean it can't build on xenial without backports.

I can take out the upgrade from debhelper 10 => 12, that should fix the travis build for older Ubuntus.

dequis:
Fun, I guess it just built fine before because it assumes 9? While changing the location of the compat file makes it assume the lowest and breaks the build. For now yeah, keep it at 10, see if that makes ubuntu precise happy again (mostly out of curiosity), and let's update that one separately.

PR #136, opened 2019-10-22
Allow to fill_by based on multiples of comma-delimited accounts/groups/protocols.

zer0def:
Allows to specify multiple comma-delimited accounts/groups/protocols in respective fields of a control channel, which when used with `fill_by` would allow to filter (and filter out) the buddy roster by more than a single entity of a given type.

As I'm not much of a `bee_group` user, I've only padded out the respective functionality for inviting and kicking users from group-based channels, so hopefully that works.

### DIFF
--- a/irc.h
+++ b/irc.h
@@ -226,9 +226,9 @@ typedef enum {
 
 struct irc_control_channel {
 	irc_control_channel_type_t type;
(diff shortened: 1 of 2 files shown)


zer0def:
Closing due to targeting the wrong branch.

PR #137, opened 2019-10-23
Allow setting multiple accounts/groups/protocols to filter a control channel's buddy roster through `fill_by`.

zer0def:
Recreation of #136 , but based in `develop`.

Allows to specify multiple comma-delimited accounts/groups/protocols in respective fields of a control channel, which when used with fill_by would allow to filter (and filter out) the buddy roster by more than a single entity of a given type.

As I'm not much of a bee_group user, I've only padded out the respective functionality for inviting and kicking users from group-based channels, so hopefully that works.


digitalcircuit:
Might be worthwhile updating the documentation to make note of this, too: https://github.com/bitlbee/bitlbee/blob/e5d2c5607310ef9669a051a1b93e019319495955/doc/user-guide/commands.xml#L929

irc_channel.c
```diff
@@ -956,20 +980,30 @@ gboolean irc_channel_wants_user(irc_channel_t *ic, irc_user_t *iu)
...
 		return FALSE;
 	}
 
 	switch (icc->type & IRC_CC_TYPE_MASK) {
 	case IRC_CC_TYPE_GROUP:
-		ret = iu->bu->group == icc->group;
+		for (l = icc->groups; l; l = l->next) {
+			ret = iu->bu->group == l->data;
+			if (ret) break;
+		}
```
dequis: `g_slist_find`?

irc_channel.c
```diff
@@ -803,6 +803,7 @@ static gboolean control_channel_init(irc_channel_t *ic)
 
 	ic->data = icc = g_new0(struct irc_control_channel, 1);
 	icc->type = IRC_CC_TYPE_DEFAULT;
+	icc->accounts = icc->groups = icc->protocols = NULL;
```
dequis: This isn't needed, as it's being initialized with `g_new0`

irc_channel.c
```diff
@@ -760,7 +760,7 @@ static gboolean control_channel_invite(irc_channel_t *ic, irc_user_t *iu)
 	}
 
 	bu->ic->acc->prpl->add_buddy(bu->ic, bu->handle,
-	                             icc->group ? icc->group->name : NULL);
+	                             icc->groups ? ((struct bee_group *)icc->groups->data)->name : NULL);
```
dequis: You could move this to a different line, and assign `->data` to a variable of the intended type to save yourself the cast. Same below in the kick function

PR #138, opened 2019-10-27
Fix compiler warning

kensanata:
Fix "ISO C does not permit named variadic macros" when compiling under -Wvariadic-macros.

[On StackOverflow](https://stackoverflow.com/questions/6750512/gcc-warning-iso-c-does-not-permit-named-variadic-macros), Matthew Slattery  says:

> This is a [GCC extension](http://gcc.gnu.org/onlinedocs/gcc-4.2.1/gcc/Variadic-Macros.html).
> [C99](http://www.open-std.org/JTC1/SC22/WG14/www/docs/n1256.pdf) does define a way of passing variable arguments to macros (see §6.10.3/12 and §6.10.3.1/2): the variable arguments are unnamed on the left-hand side of the definitions (i.e. just ...), and referenced on the right-hand side as __VA_ARGS__


dequis:
Thanks! Enjoy your tshirt!

kensanata:
I do enjoy my T-shirts. Are you offering to send me another one? 🤔

PR #139, opened 2019-10-31
Send travis to the future

dequis:
* Update from ubuntu 12.04 to 14.04
* Remove the hack that downloads libotr from my server, which is often down. Terrible hosting, but support gave me enough credits for downtime that it's practically free now.
* Fix dpkg-buildpackage which was failing since the switch to quilt format

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: c
 script:
  - ./configure --pam=1 --ldap=1
  - make check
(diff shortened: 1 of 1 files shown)


qpfiffer:
> Remove the hack that downloads libotr from my server, which is often down. Terrible hosting, but support gave me enough credits for downtime that it's practically free now.

I love this project

PR #140, opened 2019-11-19
Fixing tests for libcheck 0.13.0

d-goldin:
Since libcheck 0.13 it's mandatory to wrap the unit-test code
in a block. This updates the tests to comply with this.

This fix is backwards compatible with libcheck 0.12.


dequis:
Ah, a rare piece of code that definitely has test coverage and I just need to let CI pass.

Thank you!

PR #142, opened 2019-11-28
g_hash_table_contains() exists since GLib >= 2.32

robert-scheck:
Usage of `g_hash_table_contains()` was introduced by commit 5c90890.


dequis:
Thanks!

PR #143, opened 2020-03-17
Make the irc test work with libevent

arkamar:
This PR fixes bug https://bugs.gentoo.org/690504 which is most probably related to https://bugs.bitlbee.org/ticket/1241.


dequis:
Thanks!

PR #144, opened 2020-04-20
Allow purple connections to set no_proxy per account.

edmcman:


protocols/purple/purple.c
```diff
@@ -364,6 +368,14 @@ static void purple_sync_settings(account_t *acc, PurpleAccount *pa)
 		const char *name = "line-auth-token";
 		purple_account_set_string(pa, name, set_getstr(&acc->set, name));
 	}
+
+        PurpleProxyInfo *proxyinfo = purple_account_get_proxy_info(pa);
+        if (!proxyinfo) {
+          proxyinfo = purple_proxy_info_new();
+        }
+        assert (proxyinfo);
```
dequis, line 376: We don't normally use assert (with break), usually just `g_return_if_fail` or similar. What are you asserting here anyway? If it's out of memory then just don't.

edmcman: Thanks, I figured that wasn't the right thing to do, but wasn't sure what to do instead.

Thaodan: Squash that with the previous commit?

dequis: I normally just do that on merge

Thaodan: What about merging this?

jelmer (changes requested):
Sorry for the delay; can you rebase this so we can get the benefit of GH actions?

edmcman:
I'll do this but I'm on vacation this week.

edmcman:
I have rebased this to the current master.  It looks like someone has to approve the GH actions.

jelmer:
> I have rebased this to the current master. It looks like someone has to approve the GH actions.

Just did so.

edmcman:
@jelmer Is there anything further?

PR #145, opened 2020-04-23
Compare tweets by status ID as fallback for identical dates.

iguanaonmystack:
This fixes an issue where a thread composed using Twitter's thread interface is displayed out of order. With this change a series of statuses in a thread fetched during one poll of the timeline are correctly arranged and marked as replies to each other, rather than being in an arbitrary (often wrong) order.

As far as I can tell this doesn't have any knock-on effects from anything relying on twitter_compare_elements returning 0.


dequis:
Thanks!

PR #146, opened 2020-05-06
Add support for group reply using auto_populate_reply_metadata

iguanaonmystack:
With apologies to the author of #99, this patch uses the auto_populate_reply_metadata post argument to add a 'greply' (group reply) command. It seems to work in practice.

Possible future additions: a setting to make this the default 'reply' behaviour; support for the twitter API argument 'exclude_reply_user_ids' which allow given usernames to be untagged. Ref: https://developer.twitter.com/en/docs/tweets/tweet-updates


dequis:
Thanks!

PR #147, opened 2020-05-12
twitter: account_tag for nick and channel

Thaodan:
- twitter account is accessible now either under <account_tag> nickname or through #<account_tag> channel
- in contrast to status quo (twitter_<handle> / #twitter_<handle>) it  enables user to have a consistent namings, configurable by setting
   particular twitter account's tag
Avoid sending pings every time a new message is send in the twitter channel and only when the nick is nickname is mentioned.

Credit goes to @mkoskar (https://bugs.bitlbee.org/ticket/1173).
Rebased on current git

### DIFF
--- a/doc/user-guide/commands.xml
+++ b/doc/user-guide/commands.xml
@@ -51,11 +51,11 @@
 					</para>
 					
 					<para>
-						By default all your Twitter contacts will appear in a new channel called #twitter_yourusername. You can change this behaviour using the <emphasis>mode</emphasis> setting (see <emphasis>help set mode</emphasis>).
+						By default all your Twitter contacts will appear in a new channel called #&lt;account_tag&gt;. You can change this behaviour using the <emphasis>mode</emphasis> setting (see <emphasis>help set mode</emphasis>).
 					</para>
 					
 					<para>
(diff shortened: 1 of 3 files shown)


dequis:
As I wrote there:

>While the idea behind the patch is good, there should be some way to keep previously added accounts under the old style channel names. Not entirely sure how, other than making this behavior optional under a setting. 

To reply to your comment:

> Maybe set the default tag of the twitter account to tag_<handle>? 

Changing the account's tag is a UX breaking change too.

Thaodan:
I'm looking into this patch again to have it ready for merging.

I wonder important it is that things stay the same since this would imply that both behavior would need to stay.

Setting the tag to the <nick> for twitter accounts  for new accounts would be a middle ground that doesn't break existing UX.

jelmer (changes requested):
please also rebase to enable github actions

Thaodan:
Rebased.

jelmer:
- Sorry, this just dropped off my radar

Thaodan:
Twitter support isn't working and as far as I know 3rd-party clients are against Twitter TOS.
Should we keep working on Twitter support?

jelmer:
Ah, right

PR #148, opened 2020-08-30
Fix some lintian issues

jelmer:
Fix some lintian issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ cov-int
 .gitignore
 bitlbee.service
 bitlbee@.service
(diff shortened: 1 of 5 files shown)


dequis:
I was going to complain about the compat bump but travis did that already

jelmer:
Ah, sorry. Bug in the tool since it didn't know about trusty - fixed there now (https://salsa.debian.org/jelmer/lintian-brush/-/commit/1dd3c74ea2891b6a74688bc669acd28087d7a962) & rerun.

PR #150, opened 2020-08-30
Use dh_installsystemd rather than deprecated dh-systemd

jelmer:
Use dh_installsystemd rather than deprecated dh-systemd.


debian/rules
```diff
@@ -123,4 +127,3 @@ binary-%: build install
 	DH_OPTIONS=-p$* $(MAKE) -f debian/rules binary-common
 
 binary: binary-arch
-.PHONY: build clean binary-arch binary-common binary install
```
dequis, line 126: Why this removal?

jelmer:
Already applied.

PR #153, opened 2020-11-16
Properly terminate bitlbee with systemctl

Perdu:
Fixes the bug I posted here: https://bugs.bitlbee.org/ticket/1323

Environment : archlinux

Additional plugin: bitlbee-facebook

Steps to reproduce:
- verify that 0 bitlbee processes are running
- `$ sudo systemctl start bitlbee`
- connect to bitlbee server
- observe that 3 bitlbee processes are running
- `$ sudo systemctl stop bitlbee`
- 2 bitlbee processes remain

### DIFF
--- a/init/bitlbee.service.in
+++ b/init/bitlbee.service.in
@@ -3,7 +3,7 @@ Description=BitlBee IRC/IM gateway
 
 [Service]
 ExecStart=@sbindir@bitlbee -F -n
-KillMode=process
+KillMode=control-group
 
 [Install]
 WantedBy=multi-user.target


dequis:
Yeah, uh, sorry, that's not a bug, that's a feature.

`control-group` is the default value for KillMode if that line is missing, and I added that line to introduce this behavior [in this commit](https://github.com/bitlbee/bitlbee/commit/ad5a9078ff68c2556a2b692855f04938efc06e3d).

The idea is to allow restarting the daemon during package updates etc and not kill existing sessions. `sshd` works the same way.

Perdu:
I see, but it does have the side-effect of breaking `systemctl stop` normal functioning. Could other options such as `Restart=` or `ExecReload=` achieve the same purpose without side-effect?

dequis:
No idea! It's complicated. Closing this since it's going nowhere. Sorry for that.

Thaodan:
is this a workaround for some debian thing? Systemd doesn't restart deamons on upgrading them with the package manager. Restarting bitlbee should restart the whole daemon.

Perdu:
"Environment : archlinux"

Thaodan:
> "Environment : archlinux"

The comment referred too the comment below sorry for the noise.
The point this workaround breaks restarting bitlbee using systemd, your changes fixes that.


> Yeah, uh, sorry, that's not a bug, that's a feature.
> 
> `control-group` is the default value for KillMode if that line is missing, and I added that line to introduce this behavior [in this commit](https://github.com/bitlbee/bitlbee/commit/ad5a9078ff68c2556a2b692855f04938efc06e3d).
> 
> The idea is to allow restarting the daemon during package updates etc and not kill existing sessions. `sshd` works the same way.

PR #155, opened 2021-03-12
xmpp xep-0191/blocklist

xXx-kansei-dorifto-xXx:
jabber_data now carries a feature list which is filled at login. repeated filling does not destroy duplicates, probably not an issue for now?

### DIFF
--- a/protocols/jabber/block.c
+++ b/protocols/jabber/block.c
@@ -1,0 +1,94 @@
+/***************************************************************************\
+*                                                                           *
+*  BitlBee - An IRC to IM gateway                                           *
+*  Jabber module - Handling of blocking                                     *
+*                                                                           *
+*  Copyright 2021 / <>                                                      *
+*                                                                           *
+*  This program is free software; you can redistribute it and/or modify     *
+*  it under the terms of the GNU General Public License as published by     *
+*  the Free Software Foundation; either version 2 of the License, or        *
+*  (at your option) any later version.                                      *
+*                                                                           *
+*  This program is distributed in the hope that it will be useful,          *
+*  but WITHOUT ANY WARRANTY; without even the implied warranty of           *
+*  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the            *
+*  GNU General Public License for more details.                             *
+*                                                                           *
+*  You should have received a copy of the GNU General Public License along  *
+*  with this program; if not, write to the Free Software Foundation, Inc.,  *
+*  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.              *
+*                                                                           *
+\***************************************************************************/
+
+#include "jabber.h"
+
+int jabber_block_feature(struct im_connection *ic)
+{
+	struct jabber_data *jd = ic->proto_data;
+	GSList *features = jd->features;
+	int foundbk = FALSE;
+	struct prpl *prpl = ic->acc->prpl;
+
+	while (features) {
+		if (!strcmp(features->data, XMLNS_BLOCK)) {
+			foundbk = TRUE;
+		}
+		features = g_slist_next(features);
+	}
+
+	if (foundbk) {
+		prpl->add_deny = jabber_buddy_block;
+		prpl->rem_deny = jabber_buddy_unblock;
+		prpl->add_permit = jabber_buddy_permit;
+		prpl->rem_permit = jabber_buddy_unpermit;
+	} else {
+		prpl->add_deny = NULL;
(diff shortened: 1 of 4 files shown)


protocols/jabber/block.c
```diff
@@ -0,0 +1,94 @@
...
+	GSList *features = jd->features;
+	int foundbk = FALSE;
+	struct prpl *prpl = ic->acc->prpl;
+
+	while (features) {
+		if (!strcmp(features->data, XMLNS_BLOCK)) {
+			foundbk = TRUE;
+		}
+		features = g_slist_next(features);
+	}
```
dequis, line 38: `g_slist_find_custom(jd->features, XMLNS_BLOCK, (GCompareFunc) strcmp)`

(edit: this comment is kinda made obsolete by one i wrote later, keep reading)

protocols/jabber/block.c
```diff
@@ -0,0 +1,94 @@
...
+{
+	jabber_buddy_blockunblock(ic, who, FALSE);
+}
+
+void jabber_buddy_permit(struct im_connection *ic, char *who)
+{
+}
+void jabber_buddy_unpermit(struct im_connection *ic, char *who)
+{
+}
```
dequis, line 94: Can you add a comment for these saying something about how the XEP-0191 blocking model doesn't have allow lists, so these don't need to do anything?

(also maybe turn them into a single function that does nothing? although i admit this looks neater)

protocols/jabber/iq.c
```diff
@@ -1040,6 +1051,8 @@ static xt_status jabber_iq_disco_server_response(struct im_connection *ic,
 		}
 	}
 
+	jabber_block_feature(ic);
```
dequis, line 1054: Hmm this function could use a better name, it looks weird. Does it.. block some feature? Does it feature blocking? We just don't know

protocols/jabber/block.c
```diff
@@ -0,0 +1,94 @@
...
+	if (foundbk) {
+		prpl->add_deny = jabber_buddy_block;
+		prpl->rem_deny = jabber_buddy_unblock;
+		prpl->add_permit = jabber_buddy_permit;
+		prpl->rem_permit = jabber_buddy_unpermit;
+	} else {
+		prpl->add_deny = NULL;
+		prpl->rem_deny = NULL;
+		prpl->add_permit = NULL;
+		prpl->rem_permit = NULL;
```
dequis, line 49: So these are initialized to NULL at startup anyway. Do you expect the feature to go away? If not i'd just nuke this.

protocols/jabber/iq.c
```diff
@@ -1006,12 +1006,23 @@ static xt_status jabber_iq_disco_server_response(struct im_connection *ic,
...
+	c = query->children;
+
+	while ((c = xt_find_node(c, "feature"))) {
+		feature = xt_find_attr(c, "var");
+		if (feature) {
+			jd->features = g_slist_append(jd->features, g_strdup(feature));
+		}
+		c = c->next;
+	}
+
```
dequis, line 1025: Hmm so this receives a disco response and puts every item in jd->features. If a new response arrives, there's not really any way to forget about a feature. I'm not saying there /should/ be, but the other part of the code kinda wants to set things to NULL if it's gone, so there's an answer to that question.

On the other hand: if i'm reading this right, the only purpose of saving features in this linked list is to keep it around to later attach the handlers in block.c.

Why not just use `xt_find_node_by_attr` as with the other features, and attach the handlers to the prpl here?

dequis (commented):
Neat! Just a few suggestions to simplify the thing.

xXx-kansei-dorifto-xXx:
simplified the thing :D

protocols/jabber/block.c
```diff
@@ -0,0 +1,67 @@
+/***************************************************************************\
+*                                                                           *
+*  BitlBee - An IRC to IM gateway                                           *
+*  Jabber module - Handling of blocking                                     *
+*                                                                           *
+*  Copyright 2021 / <>                                                      *
```
jelmer, line 6: This should have a proper copyright statement

jelmer (changes requested):
looks good now, one minor comment left

protocols/jabber/block.c
xXx-kansei-dorifto-xXx: how would that look like? we just copypasted it from some other file and replaced name/mail with our git details, thought that would be fine.

jelmer: This particular line is just empty, it should have your identity as the copyright owner.

PR #157, opened 2021-05-04
Accomodate filenames containing spaces in a DCC transfer.

vjorlikowski:
Many (most?) IRC clients accomodate receiving files that contain spaces in their names by having quotes surround the file name.

This quick fix allows bitlbee to handle that case.

An alternative implementation option would be to allow selecting either quotes surrounding the filename, or replacing all spaces in the offered file name using underbar characters.

File names containing quotes should probably receive additional processing, in order to ensure the quotes internal to the filename are properly escaped.

This patch, however, allows me to exchange files containing spaces in the names with a colleague who uses XMPP via Adium, with my client being a combination of bitlbee and ERC (under emacs).

### DIFF
--- a/dcc.c
+++ b/dcc.c
@@ -210,7 +210,7 @@ int dccs_send_request(struct dcc_file_transfer *df, irc_user_t *iu, struct socka
 
 	port = ntohs(port);
 
-	cmd = g_strdup_printf("\001DCC SEND %s %s %u %zu\001",
+	cmd = g_strdup_printf("\001DCC SEND \"%s\" %s %u %zu\001",
 	                      df->ft->file_name, ipaddr, port, df->ft->file_size);
 
 	irc_send_msg_raw(iu, "PRIVMSG", iu->irc->user->nick, cmd);


jelmer:
For compatibility with clients that don't support quotes, perhaps only add the quotes when strictly necessary?

jelmer (changes requested):
See inline comment.

Thaodan:
Which client doesn't support quotes? Adding an extra check sounds unnecessary if it is just about quoting a file.

jelmer:
> Which client doesn't support quotes? Adding an extra check sounds unnecessary if it is just about quoting a file.
How do you know all clients support quotes? There may be some out there, it's hard to know for sure - clients clearly have to support lack of quotes as well, or bitlbee wouldn't work with them today.

vjorlikowski:
Apologies for the delayed response.
How would I go about detecting a client that doesn't support quotes?

I'm willing to make the change - just not quite sure how to go about it...

jelmer:
> Apologies for the delayed response. How would I go about detecting a client that doesn't support quotes?
> 
> I'm willing to make the change - just not quite sure how to go about it...

Given the number of clients out there, I think that's almost impossible to test with certainty. Which is why I'd rather be conservative about changing this behaviour  by only sending quotes when we need to  (because the filename e.g. contains spaces). That means that if a client doesn't support double quotes, at least we only break them when they send files with spaces in the name.

jelmer:
Also, please escape e.g. quotes when quoting.

jelmer (changes requested):
See earlier comments

PR #159, opened 2022-02-11
Solaris 10 Fixes

pekdon:
This PR contains the changes I had to make to get bitlbee compiling under Solaris 10.

The most major changes are in the configure script updating it to be Solaris /bin/sh compatible, apart from that it's uint16_t over u_int16_t and a strcasestr compatability function.

If you want some but not all parts of this let me know and I'll split it up in separate PRs, creating this in go to see wheter or not you migth accept changes like this in the first place.

### DIFF
--- a/configure
+++ b/configure
@@ -927,6 +927,9 @@ SunOS )
 	echo 'EFLAGS+=-lresolv -lnsl -lsocket' >> Makefile.settings
 	echo 'STRIP=\# skip strip' >> Makefile.settings
 	echo '#define NO_FD_PASSING' >> config.h
+	if [ "$arch_rel" = "5.10" ]; then
+		echo '#define NO_STRCASESTR' >> config.h
+	fi
 ;;
 AIX )
 	echo 'EFLAGS+=-Wl,-brtl' >> Makefile.settings
--- a/configure
+++ b/configure
@@ -55,15 +55,20 @@ ldap=0
 pie=1
 
 arch=$(uname -s)
+arch_rel=$(uname -r)
 
 GLIB_MIN_VERSION=2.32
 
-# Cygwin and Darwin don't support PIC/PIE
+# Cygwin, Darwin and SunOS < 5.11 do not support PIC/PIE
 case "$arch" in
 	CYGWIN* )
 		pie=0;;
 	Darwin )
 		pie=0;;
+	SunOS )
+		if [ "$arch_rel" = "5.10" ]; then
+			pie=0;;
+		fi
 esac
 
 get_version() {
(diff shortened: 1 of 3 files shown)


lib/misc.c
```diff
@@ -792,3 +792,21 @@ guint b_istr_hash(gconstpointer v)
 
 	return h;
 }
+
+#ifdef NO_STRCASESTR
```
jelmer, line 796: isn't there some function in glib that does this that we can use instead? I can't find anything at a first glance in https://people.gnome.org/~ryanl/glib-docs/glib-String-Utility-Functions.html.

configure
```diff
@@ -24,6 +24,8 @@ ipcsocket=''
 pcdir='$prefix/lib/pkgconfig'
 systemlibdirs="/lib64 /usr/lib64 /usr/local/lib64 /lib /usr/lib /usr/local/lib"
 sysroot=''
```
jelmer, line 26: I'm curious what @dequis or @Wilm0r  thing about this change; I'm personally hesitant to avoid that  all of these modern shell constructs just to please the shell on Solaris 10. Solaris 10 is ancient - that's like trying to keep support for Debian sarge around.

There's nothing to ensure Solaris 10 compatibility stays put, and given the number of lines in configure that this touches we're very likely to break support next time we make a change to configure. And even if we remember, there's no good way to test that we don't break it.

lib/misc.c
pekdon: I did look and did not find anything, and Google does not seem to give me anything useful either.

I can have another look tomorrow, but not sure what it could be named.

configure
pekdon: I _think_ /bin/sh is as broken on Solaris 11 however I'll verify that tomorrow too, if that makes any difference.

This is of course possible to work around just doing #!/usr/bin/env sh (so /usr/xpg4/bin/sh can be used, or whatever is put in the path) or executing the script like bash ./configure,

pekdon: I was wrong with the shell on Solaris 11, the existing configure script works just fine. Judging from the Illumos man pages it should be fine on other Solaris derivates as well.

Want me to drop it from the PR?

lib/misc.c
pekdon: Didn't find anything at my second glance either.

jelmer: Okay, let's leave that for the moment.

configure
jelmer: Yeah, please do - or perhaps create a separate PR for it if you do feel strongly about it. That way we can at least land these changes and perhaps have further discussion about the shell compat.

pekdon: I've removed those changes from the PR now.

jelmer (approved):

PR #151, opened 2020-08-30
Support using system libjsonparser

jelmer:
Support using system libjsonparser.

PR #162, opened 2022-08-02
Fix last places breaking strict-aliasing rules

arkamar:
This PR fixes last two places in `lib/json.c` breaking strict-aliasing rules. The fix is backported from upstream https://github.com/json-parser/json-parser. `-fno-strict-aliasing` seems to be no longer necessary, therefore, I dropped it out of default `CFLAGS`. The fix is related to downstream bug https://bugs.gentoo.org/861371.

### DIFF
--- a/configure
+++ b/configure
@@ -311,7 +311,7 @@ if [ "$debug" = "1" ]; then
 	echo 'DEBUG=1' >> Makefile.settings
 	CFLAGS="$CFLAGS -g3 -DDEBUG -O0"
 else
-	[ -z "$CFLAGS" ] && CFLAGS="-g -O2 -fno-strict-aliasing"
+	[ -z "$CFLAGS" ] && CFLAGS="-g -O2"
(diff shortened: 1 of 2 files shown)


Thaodan:
Can't we just remove the in tree json.c and submodule json-parser for those that don't want to package it?

arkamar:
This question is more related to https://github.com/bitlbee/bitlbee/pull/151, I am just trying to resolve the last problems with strict-aliasing rule. Personally, I don't see much difference between in tree json.c and submodule json-parser from packager point of view.

Thaodan:
I see the diff as that we can just update the submodule and be done with it.
But I agree in the short run it doesn't make much of a difference.

arkamar:
Yes, I see the difference from devs point of view as well :)

jelmer (approved):

arkamar:
Thanks!

PR #164, opened 2022-11-28
Add HTML5 escape sequence for '

sc68cal:
This is based off [a bug report and discussion](https://alexschroeder.ch/software/HTML_decoding_needed_after_Mastodon_3.5.0) in the mastodon plugin for Bitlbee, where mastodon sends `&#39;` for single quotes in toots.


sc68cal:
Looks like this needs some work, since it's turning `&#39;` into ` '#39;`

jelmer:
It would probably make sense to support numbered entities across the board, rather than specifically #39.

sc68cal:
@jelmer I'm fine with adding numbered entities, but I'd like some feedback on structure. Do I just add more entries to `ent` and keep the structure as is? Or are you thinking that the `htmlentity ` struct needs changing, to hold the numeric representation?

```c
typedef struct htmlentity {
	char code[7];
	char is[3];
	char numeric[7];
} htmlentity_t;
```

sc68cal:
Ah, the numbers for the numeric representation map directly to their ASCII values, so maybe capture the whole string that represents the number, then `atoi`?

sc68cal:
I think for now i'll just add more to the list, rather than make large changes. Let me know if this is sufficient

jelmer:
Hmm. I think I've just changed my mind here. Supporting arbitrary numbers opens us up to a whole can of worms (do we allow you to insert &#0; ?).

Maybe we can just support &#39; for now, and reevaluate if we end up with a long list of numbered entities here. (also, perhaps there's a library we're already depending on that has this functionality?)

sc68cal:
That works for me.

lib/misc.c
```diff
@@ -198,7 +202,12 @@ void strip_html(char *in)
 			}
 		} else if (*in == '&') {
 			cs = ++in;
-			while (*in && g_ascii_isalpha(*in)) {
```
jelmer, line 201: Isn't it simpler to just replace g_ascii_isalpha with g_ascii_isalnum here?

sc68cal: That is what I did, but I don't know if `#` is considered numeric

jelmer: It wouldn't be, I meant g_ascii_isalnum() || *in == '#' here, but this works too.

jelmer (approved):

PR #165, opened 2022-12-04
Fixes related to external-json-parser

arkamar:
This PR fixes multiple issues for those who wish to use system `libjsonparser`. First two commits are related to external plugins, the last commit corrects which `json.h` should be used when `external-json-parser` is set. See commit messages for more info.

### DIFF
--- a/configure
+++ b/configure
@@ -867,6 +867,11 @@ CYGWIN* )
 	pkgconfiglibs="-L${libdir} -lbitlbee -no-undefined"
 esac
 
+pkgconfigrequires='glib-2.0'
+if [ "$external_json_parser" = '1' ]; then
+	pkgconfigrequires="$pkgconfigrequires json-parser"
+fi
+
 cat <<EOF >bitlbee.pc
 prefix=$prefix
 includedir=$includedir
@@ -876,7 +881,7 @@ datadir=$datadir
 
 Name: bitlbee
 Description: IRC to IM gateway
-Requires: glib-2.0
+Requires: $pkgconfigrequires
 Version: $BITLBEE_VERSION
 Libs: $pkgconfiglibs
 Cflags: -I\${includedir}
--- a/configure
+++ b/configure
@@ -329,7 +329,6 @@ fi
 echo "LDFLAGS=$LDFLAGS" >> Makefile.settings
 
 echo "CFLAGS=$CFLAGS $CPPFLAGS" >> Makefile.settings
-echo CFLAGS+=-I"${srcdir}" -I"${srcdir}"/lib -I"${srcdir}"/protocols -I. >> Makefile.settings
 
 echo CFLAGS+=-DHAVE_CONFIG_H -D_GNU_SOURCE >> Makefile.settings
 
@@ -420,10 +419,12 @@ echo "EXTERNAL_JSON_PARSER=$external_json_parser" >> Makefile.settings
 if [ "$external_json_parser" = "1" ]; then
     # shellcheck disable=SC2129
     echo "CFLAGS+=$(pkg-config --cflags json-parser)" >> Makefile.settings
+    echo "CFLAGS+=-DUSE_EXTERNAL_JSON_PARSER" >> Makefile.settings
     echo "LDFLAGS_BITLBEE+=$(pkg-config --libs json-parser)" >> Makefile.settings
     echo "LDFLAGS_TESTS+=$(pkg-config --libs json-parser)" >> Makefile.settings
 fi
 
+echo CFLAGS+=-I"${srcdir}" -I"${srcdir}"/lib -I"${srcdir}"/protocols -I. >> Makefile.settings
 
 detect_gnutls()
 {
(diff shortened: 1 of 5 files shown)


configure
```diff
@@ -856,7 +862,7 @@ datadir=$datadir
 
 Name: bitlbee
 Description: IRC to IM gateway
-Requires: glib-2.0
+Requires: $pkgconfigrequires
```
jelmer, line 865: Why do we need this ? AFAIK bitlbee's public API doesn't include json.h, and we don't install static libraries?

arkamar: `json_util.h` includes `json.h` directly which is part of public API. Actaully, `json_util.h` is used in `bitlbee-steam` in `steam-json.h` which suffers with
```
<root> steam - Login error: Parser: Memory allocation failure
```
when `external-json-parser` is used.

jelmer: Hmm, looking at the Makefile we currently install all headers. We should probably limit it to the specific things that e.g. plugins need.

arkamar: Yes, I agree, but in some other PR in my opinion. Anyway, `json_util.h` is currently required by `bitlbee-steam` plugin, so, this one should be probably preserved. It uses types from `json.h` in function prototypes, which somewhat supports this new requirement :)

arkamar:
Would it be possible to merge this? It would be helpful for Gentoo package.

configure
jelmer: Okay, we should definitely discourage them from using it. But let's leave that to a separate PR.

jelmer (changes requested):
please take a look at the failing tests?

arkamar:
Merge conflict was incorrectly resolved. I have rebased the branch on top of master.

jelmer (approved):

arkamar:
Thanks!

msva:
Unfortunatelly, those changes leads to build failures here, as json.h somewhy gets included from `./lib/` instead of from `/usr/include/json-parser/` :man_shrugging: 
[build.log](https://github.com/bitlbee/bitlbee/files/11159058/build.log)

arkamar:
changes from this PR are ok, I tested them in live ebuild. The failure is most probably caused by improper revert in commit 59c9fa4d3fa235d2f884. The problem is that `configure` must append `CFLAGS` lines to `Makefile.settings` in correct order because order of `-I` flags is important for the compiler.

Which does not happen right now in commit 92a03a03
```
x86_64-gentoo-linux-musl-gcc -c -march=native -O2 -pipe -frecord-gcc-switches -flto -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing -I/var/tmp/portage/net-im/bitlbee-9999/work/bitlbee-9999 -I/var/tmp/portage/net-im/bitlbee-99
99/work/bitlbee-9999/lib -I/var/tmp/portage/net-im/bitlbee-9999/work/bitlbee-9999/protocols -I. -DHAVE_CONFIG_H -D_GNU_SOURCE -Wall -Wformat -Werror=format-security -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include -pthread  -I/usr/inclu
de/json-parser  -DUSE_EXTERNAL_JSON_PARSER -I/var/tmp/portage/net-im/bitlbee-9999/work/bitlbee-9999 -I/var/tmp/portage/net-im/bitlbee-9999/work/bitlbee-9999/lib -I/var/tmp/portage/net-im/bitlbee-9999/work/bitlbee-9999/protocols -I. -MMD -M
F .depend/json_util.o.d -fPIE json_util.c -o json_util.o
In file included from json_util.h:24,
                 from json_util.c:28:
/var/tmp/portage/net-im/bitlbee-9999/work/bitlbee-9999/lib/json.h:32:2: error: #error Bitlbee was configured to use system json-parser, this header file should not be used.
   32 | #error Bitlbee was configured to use system json-parser, this header file should not be used.
      |  ^~~~~
make[1]: *** [Makefile:48: json_util.o] Error 1
make[1]: Leaving directory '/var/tmp/portage/net-im/bitlbee-9999/work/bitlbee-9999/lib'
```
See duplicated `-I/var/tmp/portage/net-im/bitlbee-9999/work/bitlbee-9999 -I/var/tmp/portage/net-im/bitlbee-9999/work/bitlbee-9999/lib -I/var/tmp/portage/net-im/bitlbee-9999/work/bitlbee-9999/protocols -I.`, first appearance should not be there.

I added the `#error` there to actually catch these situations. It is important to use `json.h` from correct place, otherwise it could result in binary incompatibility, when structures are different, for example...

robert-scheck:
I'm not sure how this shall work: `json.h` is included by BitlBee code, however the system json-parser package usually provides `json.h` in `/usr/include/json-parser/json.h` (at least on Debian and Fedora). Thus this would IMHO require in BitlBee code `<json-parser/json.h>`; just adding `-I/usr/include/json-parser` (like Gentoo does?) seems to be wrong, because there might be also other `json.h` on default include paths.

jelmer:
the system json-parser is meant to be included as <json.h>, its pkg-config adds -I/usr/include/json-parser

The best way to ensure it gets included before the bitlbee json.h is probably to add its cflags to the front of the cflags.

robert-scheck:
> pkg-config adds -I/usr/include/json-parser

Thank you for the pointer!

PR #166, opened 2023-01-30
Add support for using libmd for MD5 functions

dcantrell:
libmd is a library found on some Linux distributions that provides MD5 and other message digest functions.  It is generally found as a companion to the libbsd library.

This patch adds --external_md5=0/1/auto to let the builder select whether or not to build with the internal MD5 routines or link with libmd.

Signed-off-by: David Cantrell <dcantrell@redhat.com>

### DIFF
--- a/configure
+++ b/configure
@@ -49,6 +49,7 @@ otr=0
 
 events=glib
 external_json_parser=auto
+external_md5=auto
 ssl=auto
 
 pam=0
@@ -160,6 +161,7 @@ Option		Description				Default
 --ssl=...	SSL library to use (gnutls, nss, openssl, auto)
 							$ssl
 --external_json_parser=0/1/auto	Use External JSON parser $external_json_parser
+--external_md5=0/1/auto	Use External MD5 routines $external_md5
 --systemd=0/1  Enable/disable systemd support $systemd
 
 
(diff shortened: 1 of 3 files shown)


arkamar (commented):
This PR will suffer with simillar issue like #165, because `md5.h` is currently installed with other header files which can break external plugins if they are using it as a part of bitlbee api or if they indirectly depend on it in which case `bitlbee.pc` needs to be updated accrodingly.

jelmer:
bitlbee already uses glib; could we just (unconditionally) use its MD5 functions instead? https://docs.gtk.org/glib/struct.Checksum.html

dequis:
We already use gchecksum, that's literally what our md5 implementation is https://github.com/bitlbee/bitlbee/blob/master/lib/md5.c

jelmer:
Ah, I hadn't seen that. Do we need the extra layer of indirection though?

dequis:
It started like that, and removing it is an API breaking change, plugins may use it.

jelmer:
Yeah, although it would be nice to limit the public API to things that are actually bitlbee-related at some point - though perhaps more as a concerted effort than bit by bit.

Looks like it's actually my fault: e506d6ce40dd357e29a7c856ab2b664df69cf015

dcantrell:
Going with glib sounds like the better option to me.  I did not know how far in to glib the project wanted to go, which is why I went down the libmd path.  Anyways, I see the PR for the glib change.  I see no reason to keep this one open.

PR #168, opened 2023-01-30
Use g_memdup2() with glib versions >= 2.68.0

dcantrell:
g_memdup() was deprecated in glib 2.68.0.  The new function is g_memdup2().  Still support building with older versions of glib via macros.

Signed-off-by: David Cantrell <dcantrell@redhat.com>

### DIFF
--- a/bitlbee.h
+++ b/bitlbee.h
@@ -103,6 +103,13 @@ extern "C" {
 #define g_strcasecmp g_ascii_strcasecmp
 #define g_strncasecmp g_ascii_strncasecmp
 
+/* g_memdup() deprecated as of glib 2.68.0 */
+#ifdef GLIB_VERSION_2_68
+#define G_MEMDUP g_memdup2
+#else
+#define G_MEMDUP g_memdup
+#endif
+
 #ifndef G_GNUC_MALLOC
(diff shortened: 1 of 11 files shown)


bitlbee.h
```diff
@@ -103,6 +103,13 @@ extern "C" {
 #define g_strcasecmp g_ascii_strcasecmp
 #define g_strncasecmp g_ascii_strncasecmp
 
+/* g_memdup() deprecated as of glib 2.68.0 */
+#ifdef GLIB_VERSION_2_68
```
jelmer, line 107: rather than defining a custom macro, could we just use g_memdup2 and define that as g_memdup on anything where glib < 2.68?

dcantrell: Yeah, that's fine.  Didn't know what the favored style was in this project or how far back you wanted to support in glib.  I will update the PR.

jelmer: Please also rebase - we've now got github actions set up.

jelmer (changes requested):
just marking as changes requested

bitlbee.h
jelmer: Done.

PR #169, opened 2023-02-01
Migrate internal users of md5.h to using GChecksum directly

jelmer:
Migrate internal users of md5.h to using GChecksum directly.

Mark md5.h functions as deprecated, to encourage users to migrate to
alternatives. This seems useful even if we don't have concrete plans to
remove md5.h at this point.


dequis:
Is this worth the hassle?

jelmer:
long-term, yes, I believe this is worth it. We'll be able to remove md5.{c,h}, but will also have a smaller public API - meaning it'll be easier to inadvertently break plugins and other API users. 

I want to do the same thing for e.g. json.h. https://github.com/bitlbee/bitlbee/pull/165 becomes simpler if we wouldn't provide a JSON API at all.

jelmer:
@Wilm0r @dequis if you don't have any objections, I'll land this in a week or so.

PR #171, opened 2023-02-02
Migrate from travis to github actions

jelmer:
This migrates coverity as well, but I haven't test it and we'll have to set the appropriate secret in GitHub to make that work.


dequis (approved):
Thank you!

I added the coverity scan token to secrets

PR #174, opened 2023-03-05
Improve control over automatic properties of Systemd configuration

arkamar:
PR #163 introduced changes which would make the Gentoo package unnecessarily more complicated.

We don't use `sysusersdir` because we have different mechanism for user and group management and there is no need to force users to install `bitlbee.sysusers` file. And we are setting `--systemdsystemunitdir` in configuration phase to some concrete path. This makes the buildtime dependency on `$PKG_CONFIG systemd` unnecessary for us.

In this PR I propose to give packagers more control over systemd configuration. With this change we will be able to use
```
configure --systemd=0 --systemdsystemunitdir=/our/path
```
to fulfill our needs without introduction of new conditional dependency on systemd.

I have also unified indentation and improved description of `--systemd` option which controls only autodetection.

### DIFF
--- a/configure
+++ b/configure
@@ -776,8 +776,12 @@ if [ "$systemd" = "1" ]; then
 	if [ -z "$sysusersdir" ] ; then
 		sysusersdir=$(pkgconf_systemd_var sysusersdir sysusers_dir)
 	fi
+fi
 
+if [ -n "$systemdsystemunitdir" ]; then
 	echo "SYSTEMDSYSTEMUNITDIR=$systemdsystemunitdir" >> Makefile.settings
+fi
+if [ -n "$sysusersdir" ]; then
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -159,7 +159,8 @@ Option		Description				Default
 --ssl=...	SSL library to use (gnutls, nss, openssl, auto)
 							$ssl
 --external_json_parser=0/1/auto	Use External JSON parser $external_json_parser
---systemd=0/1  Enable/disable systemd support $systemd
+--systemd=0/1	Enable/disable automatic detection	$systemd
```
jelmer, line 162: The $systemd should probably be on the next line?

arkamar: Probably yes, I have changed it.

jelmer (approved):

PR #178, opened 2023-04-01
Build with bundled jsonparser

jelmer:


robert-scheck:
I would like to support this if you let me know which way is preferred: With bundled jsonparser…or with system one?

jelmer:
> I would like to support this if you let me know which way is preferred: With bundled jsonparser…or with system one?

This PR was meant to fix the build on GH actions, which currently fails. Unfortunately we can't just disable the build with system jsonparser as that breaks the Debian package.

robert-scheck:
Well, what is the desired outcome for the Debian package? Bundled or system jsonparser?

jelmer:
it should use the system jsonparser

PR #183, opened 2023-04-06
BitlBee 3.7 Release Proposal

arkamar:
I am proposing let's release new version. It is more than 4 years since last release.


dequis:
Wish releases were this simple!

PR #185, opened 2023-04-07
Switch from 'which' to 'command -v' to increase portability

robert-scheck:
In minimal buildroots `which` might not be installed, while `command -v` is a shell built-in in all POSIX 2008 compliant shells, supported by at least ash, bash, busybox, dash, ksh, mksh and zsh.


arkamar (approved):
LGTM

PR #186, opened 2023-04-07
jabber: fix regression with gchecksum in jabber_cache_add

dequis:
This crashed with "The checksum '%s' has been closed and cannot be updated anymore." (which is also a user after free)

Issue introduced in 1bdc6696aefb617873b56001f81d649d6ca3aa8e which is not in any released version.

There was this md5_digest_keep() function that freed a copy of the checksum instead of the one you pass. The code that it was replaced with ignored this distinction, and since jabber_cache_add() needs to reuse the GChecksum this led to this issue.

No other place in the bitlbee code used this function.


jelmer:
Oops - thanks for catching that.

jelmer (approved):

PR #187, opened 2023-04-07
twitter: switch to an app that didn't get murdered yet

dequis:
Wonder how long this will last, probably not very.

### DIFF
--- a/protocols/twitter/twitter.c
+++ b/protocols/twitter/twitter.c
@@ -355,8 +355,8 @@ static const struct oauth_service twitter_oauth = {
 	"https://api.twitter.com/oauth/request_token",
 	"https://api.twitter.com/oauth/access_token",
 	"https://api.twitter.com/oauth/authorize",
-	.consumer_key = "xsDNKJuNZYkZyMcu914uEA",
-	.consumer_secret = "FCxqcr0pXKzsF9ajmP57S3VQ8V6Drk4o2QYtqMcOszo",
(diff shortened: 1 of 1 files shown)


dequis:
image

very legit

jelmer (approved):

ilf:
Thanks for this!

Is this adjustable without changing it in the source and recompiling?

Would this warrant a release, like 3.6.1 or 3.7?

Thanks!

jelmer:
> Thanks for this!
> 
> Is this adjustable without changing it in the source and recompiling?

It's not currently adjustable, but I suppose we could add a separate setting for it? Having to create your own app isn't a great experience, though.

ilf:
Looking at other Apps and Twitters erratic behavior, it looks like users might have to change App keys more frequently in the near future and maybe use keys that are not shared widely. So an option to change this without editing the source and recompiling sure would be useful.

In the medium future, Twitter will probably just kill all third party apps and keys, though.

Wilm0r:
Yeah, I guess this will need to be a setting from now on (and leave it up to the user whether to request a key or try out one of the official-app ones, if that's still possible?), and we'll see how many weeks/months before we can just delete the module altogether.

It's nice that at least Twitter became mostly irrelevant already *before* locking all the doors now.

dequis:
>try out one of the official-app ones, if that's still possible?

It's not, otherwise i would have used one of those in this commit.

PR #188, opened 2023-04-08
Fix ./configure: line 760: [: auto: integer expression expected

robert-scheck:
Without this change the following error is raised:

```
$ ./configure 
BitlBee configure
./configure: line 760: [: auto: integer expression expected
[…]
```

And additionally some more quoting is applied.

### DIFF
--- a/configure
+++ b/configure
@@ -755,33 +755,33 @@ else
 	fi;
 fi
 
-if [ ! "$systemd" = 0 ]; then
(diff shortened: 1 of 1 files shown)


jelmer:
Which platform/shell is this?

robert-scheck:
bash-5.1.16-3.fc37.x86_64, but `$systemd -eq 1` should always raise a message because `-eq` is for numeric comparisions.

jelmer (approved):

PR #167, opened 2023-01-30
Fix up a lot of shell errors in the configure script

dcantrell:
I ran in to a number of parse errors and problems, so I spent time fixing up the configure script.

Signed-off-by: David Cantrell <dcantrell@redhat.com>

PR #189, opened 2023-04-09
configure: address ShellCheck reports and add ShellCheck job

robert-scheck:
Attempt to finally address https://github.com/bitlbee/bitlbee/commit/2bd8f392a0d066f8f3c3f937ad2e6c1d90da78b0 from https://github.com/bitlbee/bitlbee/pull/167 and https://github.com/bitlbee/bitlbee/commit/59c9fa4d3fa235d2f884e6c1a8110f9ef1d4bd25 from https://github.com/bitlbee/bitlbee/pull/176 while (hopefully) ensuring the future code quality level using ShellCheck.

Also try to slightly unify code style (indention, whitespaces) and remove superfluous semicolon.

I also tested out-of-tree building…no easter eggs?


configure
```diff
@@ -538,13 +542,12 @@ detect_resolv_dynamic()
 	esac
 	TMPFILE=$(mktemp /tmp/bitlbee-configure.XXXXXX)
 	ret=1
-	echo "$RESOLV_TESTCODE" | $CC -o $TMPFILE -x c - $LIBRESOLV >/dev/null 2>/dev/null
-	if [ "$?" = "0" ]; then
+	if echo "$RESOLV_TESTCODE" | $CC -o "$TMPFILE" -x c - $LIBRESOLV >/dev/null 2>/dev/null; then
```
jelmer, line 545: Why this change?

configure
```diff
@@ -87,28 +88,27 @@ get_version() {
 
 	BITLBEE_VERSION=$REAL_BITLBEE_VERSION
 
-	if [ -d $srcdir/.git ] && type git > /dev/null 2> /dev/null; then
-		timestamp=$(cd $srcdir; git show -s --format=%ci HEAD | sed 's/ .*$//; s/-//g')
-		branch=$(cd $srcdir; git rev-parse --abbrev-ref HEAD)
+	if [ -d "$srcdir/.git" ] && type git > /dev/null 2> /dev/null; then
+		timestamp=$(cd "$srcdir" || exit; git show -s --format=%ci HEAD | sed 's/ .*$//; s/-//g')
```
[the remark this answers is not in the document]
robert-scheck: This is [SC2164](https://www.shellcheck.net/wiki/SC2164), but would `git` commands executed in the wrong directory be better? Given the check in line 91, the `exit` in line 92 should only become relevant in race conditions, no?

configure
robert-scheck: This is [SC2181](https://www.shellcheck.net/wiki/SC2181) and should avoid some redundancy.

configure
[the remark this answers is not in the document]
jelmer: I'm not a fan of adding this sort of thing just to shut up a linter, when there's no functional advantage. In fact, the opposite since this silences the problem rather than giving you a somewhat confusing git error.

configure
jelmer: In this particular instance, I personally find the original code easier to read spread out over two lines.

configure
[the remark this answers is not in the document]
robert-scheck: It shouldn't silence the issue, because `cd` still yells at `STDERR`, but I can revert this indeed. Just to clarify: All occurrences of ` || exit`, that I added?

configure
robert-scheck: Okay, while that's likely a matter of taste, I can change it back. And all occurrences of changed `$…TESTCODE`, I guess?

configure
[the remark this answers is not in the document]
robert-scheck: If you let me know, I'm happy to update my pull request.

configure
robert-scheck: If you let me know, I'm happy to update my pull request.

jelmer: Yeah, can you perhaps revert this for now - in the interest of being conservative? I think the line is long enough to be split up.

configure
```diff
@@ -553,16 +558,17 @@ detect_resolv_static()
 	TMPFILE=$(mktemp /tmp/bitlbee-configure.XXXXXX)
 	ret=1
 	for i in $systemlibdirs; do
-		if [ -f $i/libresolv.a ]; then
-			echo "$RESOLV_TESTCODE" | $CC -o $TMPFILE -x c - -Wl,$i/libresolv.a >/dev/null 2>/dev/null
+		if [ -f "$i/libresolv.a" ]; then
+			echo "$RESOLV_TESTCODE" | $CC -o "$TMPFILE" -x c - -Wl,"$i/libresolv.a" >/dev/null 2>/dev/null
+			# shellcheck disable=SC2181
```
jelmer, line 563: can we just globally disable these, rather than having to sprinkle them around the code wherever they appear?

robert-scheck: Yes, changed.

jelmer (approved):

PR #191, opened 2023-09-12
Rename /OPER command /oper everywhere

Perdu:
Weechat is in the process of making its identifiers case-sensitive (see https://specs.weechat.org/specs/2023-001-case-sensitive-identifiers.html#changes). As a result, /OPER is no longer a valid weechat command.

This patch is an example patch to address this issue, replacing all /OPER references to /oper.

I don't know if this affects other IRC clients and if there is a standard favoring /OPER or /oper.

It's an example patch, you may want to replace the message with e.g. "use /oper or /OPER in older weechat versions".

### DIFF
--- a/bitlbee.h
+++ b/bitlbee.h
@@ -132,7 +132,7 @@ extern "C" {
 #define CONF_FILE_DEF ETCDIR "bitlbee.conf"
 
 /* Hack to give a little bit more password security on IRC: If an account has
-   this password set, use /OPER to change it. */
+   this password set, use /oper to change it. */
 #define PASSWORD_PENDING "\r\rchangeme\r\r"
 
 #include "bee.h"
--- a/doc/user-guide/commands.xml
+++ b/doc/user-guide/commands.xml
@@ -24,7 +24,7 @@
 				</para>
 
 				<para>
-					You can omit the password and enter it separately using the IRC /OPER command. This lets you enter your password without your IRC client echoing it on screen or recording it in logs.
+					You can omit the password and enter it separately using the IRC /oper command. This lets you enter your password without your IRC client echoing it on screen or recording it in logs.
 				</para>
 			</description>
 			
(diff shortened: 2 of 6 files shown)


jelmer (commented):
Hmm, not super convinced about this. I think if we change the casing of /OPER we should change the casing of other commands (/QUIT, /PART, /JOIN, etc) as well for consistency.

Even if most clients support case-sensitive commands, it's more common to write them in uppercase. The protocol level commands in the RFC are also in uppercase.

jelmer:
Why are they doing this? It's not like they'll be able to reuse "/OPER" for different purposes than "/oper" so it seems like dropping support for the uppercase version will just make things more confusing for their users.

jelmer:
I'm open to the idea of just changing /OPER to /oper, but we should make sure we don't just change /OPER but other commands in the documentation (/PART, /MSG, /JOIN, /QUIT, etc) as well. At the moment we use a mix of uppercase and lowercase in our docs.
(changing to lowercase is going to fix weechat, but who knows when the next client comes along that only accepts uppercase commands; that's merely theoretical at this point so let's not worry about it)

That said, I don't want to add "(or /oper on weechat)" to the docs - we don't want to have to document special cases for every client that doesn't follow the common conventions. weechat are the ones intentionally breaking the status quo here. That's their prerogative, but their users are going to have to deal with the fact that lots of other places on the web reference uppercase commands anyway.

dequis:
Sounds like they made bad decisions! I don't think that's our problem!

dgw:
Yeah, the `<letter>` pseudo-BNF token is defined (on [page 9 of RFC 1459](https://www.rfc-editor.org/rfc/rfc1459.html#page-9)) as either upper- or lowercase, to boot. The protocol explicitly doesn't care about the casing of commands; `OPER`, `oper`, `oPer`, `Oper`, `OpEr`, and every other combination of casings is valid.

It's silly to force client commands to be case-sensitive, but at the very least a single client's silliness shouldn't affect how BitlBee documents stuff for everyone else.

Perdu:
So, there are more opinions towards not changing anything? It's not worth crafting a better patch, then?

Perdu:
Closing, then

PR #198, opened 2025-10-05
Drop Debian packaging

jelmer:
bitlbee has been moved to the QA team in Debian, managed on salsa:

https://salsa.debian.org/debian/bitlbee

See also https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1114687

Having a copy in the upstream repository as well is confusing, so let's drop it htere.


Wilm0r:
Yeah agreed, it's about time to clean this up..

PR #199, opened 2026-08-01
Support port range in ft_listen

yourealwaysbe:
Allows multiple files to be received at once (on different ports).

Useful with purple-whatsmeow when someone sends 10 photos at once on WhatsApp.

I've been using this for over a year now without issue, so i figure it's ok for a pull request :)

### DIFF
--- a/lib/ftutil.c
+++ b/lib/ftutil.c
@@ -138,11 +138,9 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
 		}
 
 		if ((fd = socket(saddr->ss_family, SOCK_STREAM, 0)) == -1) {
-			if (current_port == port_end) {
-				g_snprintf(errmsg, sizeof(errmsg), "Opening socket: %s", strerror(errno));
-				return -1;
-			}
-			continue;
+			// trying a different port won't help
+			g_snprintf(errmsg, sizeof(errmsg), "Opening socket: %s", strerror(errno));
+			return -1;
 		}
 
 		if (bind(fd, (struct sockaddr *) saddr, saddrlen) == -1) {
--- a/lib/ftutil.c
+++ b/lib/ftutil.c
@@ -176,7 +176,7 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
 			memcpy(saddr, rp->ai_addr, saddrlen);
 			freeaddrinfo(rp);
 		} else {
-			// re-use previous address in saddr with new port
+			/* re-use previous address in saddr with new port */
 			if (saddr->ss_family == AF_INET) {
 				((struct sockaddr_in *) saddr)->sin_port = htons(current_port);
 			} else {
--- a/lib/ftutil.c
+++ b/lib/ftutil.c
@@ -86,6 +86,9 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
 	char port_range[MAX_PORT_RANGE_LEN + 1] = {0};
 	int port_start = 0, port_end = 0;
 	int current_port, port_count, port_base_offset, port_offset;
+	/* when trying a port range start at a different position each time
+	 * to reduce collisions */
+	static int port_search_start_offset = 0;
 	char *dash_pos;
 
 	if (errptr) {
@@ -155,9 +158,11 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
 	hints.ai_flags = AI_NUMERICSERV;
 
 	/* Try each port in the range
-	 * Start from a random offset to reduce collisions. */
+	 * Start from a different offset each time to reduce collisions. */
 	port_count = port_end - port_start + 1;
-	port_base_offset = rand() % port_count;
+	port_base_offset = port_search_start_offset % port_count;
+	/* Next time, start from next port along */
+	port_search_start_offset = (port_search_start_offset + 1) % (MAX_PORT + 1);
 	for (port_offset = 0; port_offset < port_count; port_offset++) {
 		current_port = port_start + (port_base_offset + port_offset) % port_count;
 		g_snprintf(port, 6, "%d", current_port);
--- a/lib/ftutil.c
+++ b/lib/ftutil.c
@@ -32,6 +32,45 @@
 		g_snprintf(errmsg, sizeof(errmsg), msg ": %s", strerror(errno)); \
 		return -1; }
 
+#define MAX_PORT 65535
+/* two port numbers or service names up to size 20, plus a - */
+#define MAX_PORT_RANGE_LEN 41
+
+/*
+ * Parse the port number in value
+ *
+ * Converts to the integer value if in range and the whole string is an
+ * integer. Else tries to look up the value by service name.
+ *
+ * Returns -1 if no valid port could be determined.
+ */
+static int parse_port(char *value) {
+	size_t len;
+	char *endptr;
+	long port;
+
+	len = strlen(value);
+	if (len == 0) {
+		return -1;
+	}
+
+	port = strtol(value, &endptr, 10);
+
+	/* if not whole string was a number */
+	if ((size_t) (endptr - value) < len) {
+		struct servent *s = getservbyname(value, NULL);
+		if (s) {
+			return ntohs(s->s_port);
+		} else {
+			return -1;
+		}
+	} else if (port < 0 || port > MAX_PORT) {
+		return -1;
+	} else {
+		return (int) port;
+	}
+}
+
 /*
  * Creates a listening socket and returns it in saddr_ptr.
  */
@@ -44,7 +83,7 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
 	struct sockaddr_storage saddrs = {0}, *saddr = &saddrs;
 	static char errmsg[1024];
 	char *ftlisten = global.conf->ft_listen;
-	char port_range[12] = {0};
+	char port_range[MAX_PORT_RANGE_LEN + 1] = {0};
 	int port_start = 0, port_end = 0;
 	int current_port, port_count, port_base_offset, port_offset;
 	char *dash_pos;
@@ -78,23 +117,23 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
 
 		if ((colon = strchr(host, ':'))) {
 			*colon = '\0';
-			strncpy(port_range, colon + 1, 11);
+			strncpy(port_range, colon + 1, MAX_PORT_RANGE_LEN);
 
 			/* Check if port is a range (contains '-') */
 			if ((dash_pos = strchr(port_range, '-'))) {
 				*dash_pos = '\0';
-				port_start = atoi(port_range);
-				port_end = atoi(dash_pos + 1);
+				port_start = parse_port(port_range);
+				port_end = parse_port(dash_pos + 1);
 				*dash_pos = '-';  /* restore for potential error messages */
 
-				if (port_start <= 0 || port_end <= 0 || port_start > port_end || port_end > 65535) {
+				if (port_start < 0 || port_end < 0 || port_start > port_end) {
 					sprintf(errmsg, "Invalid port range: %s", port_range);
 					return -1;
 				}
 			} else {
 				/* Single port */
-				port_start = port_end = atoi(port_range);
-				if (port_start <= 0 || port_start > 65535) {
+				port_start = port_end = parse_port(port_range);
+				if (port_start < 0) {
 					sprintf(errmsg, "Invalid port: %s", port_range);
 					return -1;
 				}
--- a/lib/ftutil.c
+++ b/lib/ftutil.c
@@ -44,6 +44,9 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
 	struct sockaddr_storage saddrs = {0}, *saddr = &saddrs;
 	static char errmsg[1024];
 	char *ftlisten = global.conf->ft_listen;
+	char port_range[11] = {0};
+	int port_start = 0, port_end = 0, current_port;
+	char *dash_pos;
 
 	if (errptr) {
 		*errptr = errmsg;
@@ -54,6 +57,7 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
 	/* Format is <IP-A>[:<Port-A>];<IP-B>[:<Port-B>] where
 	 * A is for connections with the bitlbee client (DCC)
 	 * and B is for connections with IM peers.
+	 * Port-A and Port-B can be ranges like "3000-3010"
 	 */
 	if (ftlisten) {
 		char *scolon = strchr(ftlisten, ';');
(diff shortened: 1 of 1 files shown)


lib/ftutil.c
```diff
@@ -44,6 +44,9 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
 	struct sockaddr_storage saddrs = {0}, *saddr = &saddrs;
 	static char errmsg[1024];
 	char *ftlisten = global.conf->ft_listen;
+	char port_range[11] = {0};
```
jelmer: why is 11 sufficient here?

lib/ftutil.c
```diff
@@ -54,6 +57,7 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
 	/* Format is <IP-A>[:<Port-A>];<IP-B>[:<Port-B>] where
 	 * A is for connections with the bitlbee client (DCC)
 	 * and B is for connections with IM peers.
+	 * Port-A and Port-B can be ranges like "3000-3010"
 	 */
 	if (ftlisten) {
```
jelmer, line 105: how do port_start and port_end get initialized if ftlisten is not set?

jelmer (changes requested):

lib/ftutil.c
```diff
@@ -90,20 +114,50 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
...
+		memcpy(saddr, rp->ai_addr, saddrlen);
+		freeaddrinfo(rp);
 
-	freeaddrinfo(rp);
+		if ((fd = socket(saddr->ss_family, SOCK_STREAM, 0)) == -1) {
+			if (current_port == port_end) {
+				g_snprintf(errmsg, sizeof(errmsg), "Opening socket: %s", strerror(errno));
+				return -1;
+			}
+			continue;
```
jelmer: socket() failure is unrelated to the specific port number, so why do we continue the loop?

lib/ftutil.c
```diff
@@ -90,20 +114,50 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
...
+		}
 
-	ASSERTSOCKOP(fd = socket(saddr->ss_family, SOCK_STREAM, 0), "Opening socket");
-	ASSERTSOCKOP(bind(fd, ( struct sockaddr *) saddr, saddrlen), "Binding socket");
-	ASSERTSOCKOP(listen(fd, 1), "Making socket listen");
+		if (bind(fd, (struct sockaddr *) saddr, saddrlen) == -1) {
+			close(fd);
+			if (current_port == port_end) {
+				g_snprintf(errmsg, sizeof(errmsg), "Binding socket: %s", strerror(errno));
```
jelmer: this error message should probably provide some more background, i.e. mention that the entire port range was tried.

lib/ftutil.c
```diff
@@ -90,20 +114,50 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
...
-		return -1;
-	}
+	/* Try each port in the range */
+	for (current_port = port_start; current_port <= port_end; current_port++) {
+		if (current_port > 0) {
+			g_snprintf(port, 6, "%d", current_port);
+		}
 
-	saddrlen = rp->ai_addrlen;
+		if ((gret = getaddrinfo(host, port, &hints, &rp)) != 0) {
```
jelmer: host isn't going to change and port is a number so it seems wasteful to have a getaddrinfo call here.

jelmer (changes requested):

lib/ftutil.c
yourealwaysbe: Ack, sorry, it should be 12: two 5-digit numbers, a '-' and a trailing null.

lib/ftutil.c
yourealwaysbe: They're initialised to 0 when declared.

lib/ftutil.c
yourealwaysbe: Fair point, it can just return here rather than retry.

lib/ftutil.c
yourealwaysbe: Ok.

lib/ftutil.c
yourealwaysbe: Do you mean to change the port number by manipulating rp->ai_info directly? I.e. casting to a sockaddr_in and relying on sockaddr_in6 having the same layout up to the port number?

I'm not super experienced with sockets in C, so figured getaddrinfo is safe.

yourealwaysbe:
Thanks for comments, will update. I've added a few follow-up questions.

lib/ftutil.c
jelmer: right, but isn't that a change in behavour ?

lib/ftutil.c
jelmer: You're calling getaddrinfo() in a loop here, rather than only once - while the service is going to be a number anyway.

lib/ftutil.c
yourealwaysbe: I don't think so, or don't intend it to be.

On the current master branch, [port is initially set to 0 on line 52](https://github.com/bitlbee/bitlbee/blob/af09afd56f4c67c607947cde7e78d0038c3611f4/lib/ftutil.c#L52), which is used if `ftlisten` is not set.

In this request, `port_start` and `port_end` are initialised to 0, meaning that the for loop only tries port 0 if `ftlisten` is not set. The 0 is copied into `port` before `getaddrinfo` is called (edit: `port` is still set to 0 at the start, and the loop that tries available ports only changes `port` if it's trying a non-zero port).

lib/ftutil.c
yourealwaysbe: Sorry, i don't understand. The value of `port` changes between calls ([set at the start of the loop](https://github.com/bitlbee/bitlbee/pull/199/changes/BASE..0f91b508d8bcaf61f37b84937b0a8413ca2aa425#diff-3a5771a20ba4a8d953bdcf54113a6aa0c624d5a70c34807e8e863ffdd7e60e38R120)). So i thought i had to call `getaddrinfo` again to get the addresses for the updated port number.

yourealwaysbe: Poking around, i think its safe to update rp->ai_info directly with the next port number to try (branching on socket type). That would avoid multiple calls to getaddrinfo.

yourealwaysbe:
I've updated:

* Fixed `port_range` size.
* Call getaddrinfo only once.
* Fail immediately if socket can't be created.
* Give tried port range in error messages.

I also adjusted the algorithm to start from a random port in the range, then cycle. I think that should be more efficient than starting at the beginning of the range each time and getting guaranteed n^2 complexity.

lib/ftutil.c
```diff
@@ -90,20 +115,58 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
...
 	hints.ai_flags = AI_NUMERICSERV;
 
-	if ((gret = getaddrinfo(host, port, &hints, &rp)) != 0) {
-		sprintf(errmsg, "getaddrinfo() failed: %s", gai_strerror(gret));
-		return -1;
-	}
+	/* Try each port in the range
+	 * Start from a random offset to reduce collisions. */
+	port_count = port_end - port_start + 1;
+	port_base_offset = rand() % port_count;
```
jelmer: do we actually seed anywhere?

lib/ftutil.c
```diff
@@ -90,20 +115,58 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
...
+		if (port_offset == 0) {
+			if ((gret = getaddrinfo(host, port, &hints, &rp)) != 0) {
+				sprintf(errmsg, "getaddrinfo() failed: %s", gai_strerror(gret));
+				return -1;
+			}
+			saddrlen = rp->ai_addrlen;
+			memcpy(saddr, rp->ai_addr, saddrlen);
+			freeaddrinfo(rp);
+		} else {
+			// re-use previous address in saddr with new port
```
jelmer: where does saddrlen get initialized?

lib/ftutil.c
```diff
@@ -90,20 +115,58 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
...
+			if (saddr->ss_family == AF_INET) {
+				((struct sockaddr_in *) saddr)->sin_port = htons(current_port);
+			} else {
+				((struct sockaddr_in6 *) saddr)->sin6_port = htons(current_port);
+			}
+		}
 
-	saddrlen = rp->ai_addrlen;
+		if ((fd = socket(saddr->ss_family, SOCK_STREAM, 0)) == -1) {
+			// trying a different port won't help
```
jelmer: I'm not sure we officially support C99 yet?

lib/ftutil.c
```diff
@@ -73,11 +78,31 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
...
+
+			/* Check if port is a range (contains '-') */
+			if ((dash_pos = strchr(port_range, '-'))) {
+				*dash_pos = '\0';
+				port_start = atoi(port_range);
+				port_end = atoi(dash_pos + 1);
+				*dash_pos = '-';  /* restore for potential error messages */
+
+				if (port_start <= 0 || port_end <= 0 || port_start > port_end || port_end > 65535) {
+					sprintf(errmsg, "Invalid port range: %s", port_range);
```
jelmer: please use g_snprintf here as well

lib/ftutil.c
```diff
@@ -90,20 +115,58 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
...
+		if (listen(fd, 1) == -1) {
+			close(fd);
+			if (port_offset == port_count - 1) {
+				g_snprintf(errmsg, sizeof(errmsg), "Making socket listen: %s. Tried port(s) %s", strerror(errno), port_range);
+				return -1;
+			}
+			continue;
+		}
 
-	ASSERTSOCKOP(fd = socket(saddr->ss_family, SOCK_STREAM, 0), "Opening socket");
```
jelmer, line 104: is ASSERTSOCKOP still used anywhere?

lib/ftutil.c
```diff
@@ -73,11 +78,31 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
...
+				port_end = atoi(dash_pos + 1);
+				*dash_pos = '-';  /* restore for potential error messages */
+
+				if (port_start <= 0 || port_end <= 0 || port_start > port_end || port_end > 65535) {
+					sprintf(errmsg, "Invalid port range: %s", port_range);
+					return -1;
+				}
+			} else {
+				/* Single port */
+				port_start = port_end = atoi(port_range);
```
jelmer: doesn't this break on non-integer ports, i.e. service names? And what happens if port_range is not a valid integer?

lib/ftutil.c
```diff
@@ -90,20 +115,58 @@ int ft_listen(struct sockaddr_storage *saddr_ptr, char *host, char *port, int co
...
+			}
+			continue;
+		}
 
-	ASSERTSOCKOP(fd = socket(saddr->ss_family, SOCK_STREAM, 0), "Opening socket");
-	ASSERTSOCKOP(bind(fd, ( struct sockaddr *) saddr, saddrlen), "Binding socket");
-	ASSERTSOCKOP(listen(fd, 1), "Making socket listen");
+		/* Success! Break out of the loop */
+		break;
+	}
```
jelmer, line 214: what happens if we run out of ports to try?

lib/ftutil.c
yourealwaysbe: No -- i don't think it's necessary. It's really just a way to save state between calls without a global variable (ideally you'd start at the next slot since the last call and cycle). It doesn't have to be a different sequence of numbers each time the program runs.

lib/ftutil.c
yourealwaysbe: Line 131.

lib/ftutil.c
yourealwaysbe: Ah, sorry, will fix that.

yourealwaysbe:
Poking around, it looks like i implemented ["linear probing"](https://en.wikipedia.org/wiki/Linear_probing) which appears to n^1.5 expected complexity to use up all ports*. Just picking random ports until a free one is found should have n ln n, but then you need an array or something to know when to stop. So perhaps linear probing is ok, esp since i didn't have any issues with the original n^2 version (you'd have to be sending a lot of files, at which point you probably have other issues). There are also other strategies than linear probing, but they tend to require assumptions on the number of available ports to guarantee all ports in the range can be used.

(* linear probing is about inserting into a hash table but is practically the same as searching for a free port here. With help from AI (so maybe wrong) using results from Knuth AOCP Section 6.4 Theorem K: cost of successfully searching an item when all n slots are filled is (1/2)(1 + Q0(n,n-1)) where Q0(n,n-1) is noted to be Q(n) calculated as O(n^1/2) in Volume 1 (Q0 and Q properly defined in AOCP). The expected time of a successful search is also (1/n) times the sum of the insertion costs to fill the table (since finding an item follows the same path as the original insertion). So Q(n) ~= (1/n) sum of insertions. Multiplying by n gives that the sum of the insertions is roughly n * Q(n) or n * O(n^1/2) = O(n^3/2). Maybe.. At any rate, Knuth also shows that the expected time per insertion is constant for a given proportion of slots used (assuming an arbitrarily large number of slots). Upper-bounds for 70% full is 5.5 tries, 75% is 8.5 tries, then it's downhill from there (capping out at O(n)).)

lib/ftutil.c
jelmer: Can we just avoid rand()? Using it also means we might not try all ports in the range or try some ports more than once.

jelmer (changes requested):

lib/ftutil.c
yourealwaysbe: Line 111 and Line 180. Admittedly not as useful as it was before port-ranges.

lib/ftutil.c
yourealwaysbe: Yes, good point. I didn't know about service names. Will update.

The previous code allowed up to 5 characters for the port, which is already too small for several services in my `/etc/services` file. Do you have a preferred size to allocate for service names? (20 chars? -- the longest in my services file is 15.)

lib/ftutil.c
yourealwaysbe: In the preceding code setting up the socket, if a step fails, and it's the last port number to try, the function returns -1. It only gets to the end of the loop if a port worked.

yourealwaysbe:
>Can we just avoid rand()? Using it also means we might not try all ports in the range or try some ports more than once.

(Not sure why i can't reply to this one inline.)

All ports in the range will be tried sequentially. `rand()` just picks where to start from. E.g it might try ports 6,7,8,9,10, then 1,2,3,4,5 if the range is 1-10 and 6 is the random start point.

In my initial implementation it always tried 1,2,3,4,5,6,7,8,9,10. So if ft_listen were called 5 times rapidly (e.g. when five photos get sent from WhatsApp) then the first would go on port 1, the second would try port 1, fail, then send on port 2, the third would fail on ports 1 and 2 then use port 3, and so on. Giving n^2 failures to send n files.

One solution is to live with that (seems not to cause me much trouble).

The nicest would be if the first call to ft_listen starts at port 1, the second at port 2, and so on. For that we'd need a global variable to remember the port used in the last call so it starts at the next one. I'd been avoiding global variables, but perhaps it's the best solution?

Instead of a global variable, the current version uses `rand()` to avoid starting at the same place each time.

Happy to go with whatever you prefer. (There are a number of exotic solutions to this problem it turns out.)

lib/ftutil.c
yourealwaysbe: Done, and the one after. There's still an sprintf at Line 172, but that was there before the pull request.

lib/ftutil.c
yourealwaysbe: Added a function to parse the port. First as a number. If the whole thing isn't a number, it tries `getservbyname`. This seems to match `getaddrinfo` that interprets a numeric port as a number first (even if that number is defined as a name in `/etc/services`).

I increased the buffer size for `port_range` to 42 (two 20-character port names and a - and a null).

yourealwaysbe:
I have removed `rand()` and use a static local variable instead to start from a different position in the port range each time.

As part of the same change, i stopped rejecting port 0 as an invalid port, as you can legitimately write e.g. "localhost:0" in your config.

In addition, i removed another C99 comment.

